PR #5, opened 2022-02-03
allow setting of ZYTE_API_KEY via Scrapy settings

### LINKED ISSUE #4: Allow setting of ZYTE_API_KEY via settings.py instead of os.env
`scrapy-zyte-api` uses https://github.com/zytedata/python-zyte-api which needs to have the `ZYTE_API_KEY` to be set in the Env vars. 

To follow Scrapy's conventions, we'll need to allow this to be set via the settings instead.

BurnzZ:
Fixes #4

### DIFF
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -44,8 +44,8 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
             raise NotConfigured
 
         logger.info(f"Using Zyte API Key: {zyte_api_key[:7]}")
-        os.environ["ZYTE_API_KEY"] = zyte_api_key
-        return cls(crawler.settings, crawler)
+        client = AsyncClient(api_key=zyte_api_key)
+        return cls(crawler.settings, crawler, client)
 
     def download_request(self, request: Request, spider: Spider) -> Deferred:
         if request.meta.get("zyte_api"):
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -38,8 +38,8 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         zyte_api_key = crawler.settings.get("ZYTE_API_KEY") or os.getenv("ZYTE_API_KEY")
         if not zyte_api_key:
             logger.warning(
-                f"'ZYTE_API_KEY' must be set in the spider settings or env var "
-                f"in order for ScrapyZyteAPIDownloadHandler to work."
+                "'ZYTE_API_KEY' must be set in the spider settings or env var "
+                "in order for ScrapyZyteAPIDownloadHandler to work."
             )
             raise NotConfigured
 
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -1,12 +1,13 @@
 import json
 import logging
+import os
 from base64 import b64decode
 from typing import Any, Dict, Generator, List, Optional
 
 from scrapy import Spider
 from scrapy.core.downloader.handlers.http import HTTPDownloadHandler
 from scrapy.crawler import Crawler
-from scrapy.exceptions import IgnoreRequest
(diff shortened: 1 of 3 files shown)


scrapy_zyte_api/handler.py
```diff
@@ -32,6 +33,20 @@ def __init__(
...
+        zyte_api_key = crawler.settings.get("ZYTE_API_KEY") or os.getenv("ZYTE_API_KEY")
+        if not zyte_api_key:
+            logger.warning(
+                "'ZYTE_API_KEY' must be set in the spider settings or env var "
+                "in order for ScrapyZyteAPIDownloadHandler to work."
+            )
+            raise NotConfigured
+
+        logger.info(f"Using Zyte API Key: {zyte_api_key[:7]}")
+        os.environ["ZYTE_API_KEY"] = zyte_api_key
```
kmike, line 47: What do you think about passing api_key as a parameter to AsyncClient when these is a Scrapy setting present, instead of modifying os.environ?

kmike: though if you're validating os.getenv anyways, there is no big difference

scrapy_zyte_api/handler.py
```diff
@@ -32,6 +33,20 @@ def __init__(
...
+    def from_crawler(cls, crawler):
+        zyte_api_key = crawler.settings.get("ZYTE_API_KEY") or os.getenv("ZYTE_API_KEY")
+        if not zyte_api_key:
+            logger.warning(
+                "'ZYTE_API_KEY' must be set in the spider settings or env var "
+                "in order for ScrapyZyteAPIDownloadHandler to work."
+            )
+            raise NotConfigured
+
+        logger.info(f"Using Zyte API Key: {zyte_api_key[:7]}")
```
kmike, line 46: Why is it cut? Is it for security reasons? Would it make sense to add "*" for the rest of the key?

scrapy_zyte_api/handler.py
kmike: Example where it may matter: you start 2 spiders in the same process using CrawlerRunner, and they have different API keys configured via custom_settings. In this case it won't work if you're modifying os.environ, but it'd work if you pass an API key. I know, that's really an edge case :)

scrapy_zyte_api/handler.py
BurnzZ: Yep, following the convention of `scrapy-zyte-smartproxy`: https://github.com/scrapy-plugins/scrapy-zyte-smartproxy/blob/master/scrapy_zyte_smartproxy/middleware.py#L85-L90.

I'm not quite sure about having **"*"** for the rest as it could pollute the logs a bit. Though I think having the first few chars would be enough to have a good reference on the API that's currently set.

scrapy_zyte_api/handler.py
BurnzZ: Good point! We'll handle that edge case in any case. Thanks for raising this! Updated in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/5/commits/510ae72d965994df428e1edcf18f55fa61c2e9ff

scrapy_zyte_api/handler.py
kmike: 👍 I think if wre want to change it, it'd be better to change it in both places; it's out of scope for this PR.

tests/test_api_requests.py
```diff
@@ -211,3 +212,27 @@ async def test_job_id(self, job_id):
...
             assert resp.body == f"<html>{job_id}</html>".encode("utf8")
+
+
+def test_api_key_presence():
+    from scrapy_zyte_api.handler import ScrapyZyteAPIDownloadHandler
+
+    API_KEY = "TEST_API_KEY"
+
+    # Setting the API KEY via env vars should work
+    os.environ["ZYTE_API_KEY"] = API_KEY
```
kmike, line 223: a nitpick: it'd be good to restore os.environ to its previous state on the test teardown

kmike:
Thanks @BurnzZ!

PR #10, opened 2022-04-26
introduce ZyteAPITextResponse and ZyteAPIResponse to store raw Zyte Data API Response

### LINKED ISSUE #9: Allow to access api_response
There should be a way to get complete API response. Some options:

* store it in meta
* create a custom Response class, store api response in an attribute

BurnzZ:
Fixes #9 

---

The approach of storing the Zyte API response inside the `response.meta` was avoided since it will break the expectation in Scrapy that `response.meta` comes directly from `response.request.meta`, as what [this code snippet](https://github.com/scrapy/scrapy/blob/master/scrapy/http/response/__init__.py#L65-L73) embodies:

```python
@property
def meta(self):
    try:
        return self.request.meta
    except AttributeError:
        raise AttributeError(
            "Response.meta not available, this response "
            "is not tied to any request"
        )
```

In addition, this is what the [official Scrapy docs](https://docs.scrapy.org/en/latest/topics/request-response.html?highlight=response#scrapy.http.Response.meta) says about `response.meta`:

> Unlike the `Response.request` attribute, the `Response.meta` attribute is propagated along redirects and retries, so you will get the original `Request.meta` sent from your spider.

Thus, this PR introduces these new classes which inherits from Scrapy's `Response` and `TextResponse` respectively:

- `ZyteAPITextResponse`
- `ZyteAPIResponse` 

This enables users to retrieve the raw Zyte Data API response as a `dict` via `response.zyte_api_response`.

### TODO:

- [x] more tests
- [x] update README
- [x] update CHANGELOG

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,8 +6,8 @@ TBD
 
 * Introduce ``ZyteAPIResponse`` and ``ZyteAPITextResponse`` which are subclasses
   of ``scrapy.http.Response`` and ``scrapy.http.TextResponse`` respectively.
-  These new response classes hold the raw Zyte Data API response in its
-  ``zyte_api`` attribute.
+  These new response classes hold the raw Zyte Data API response in the
+  ``raw_api_response`` attribute.
 
 0.1.0 (2022-02-03)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ TBD
 * Introduce ``ZyteAPIResponse`` and ``ZyteAPITextResponse`` which are subclasses
   of ``scrapy.http.Response`` and ``scrapy.http.TextResponse`` respectively.
   These new response classes hold the raw Zyte API response in its
-  ``zyte_api_response`` attribute.
+  ``zyte_api`` attribute.
 
 0.1.0 (2022-02-03)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Changes
 =======
 
+TBD
+---
+
+* Introduce ``ZyteAPIResponse`` and ``ZyteAPITextResponse`` which are subclasses
+  of ``scrapy.http.Response`` and ``scrapy.http.TextResponse`` respectively.
+  These new response classes hold the raw Zyte API response in its
+  ``zyte_api_response`` attribute.
+
 0.1.0 (2022-02-03)
 ------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ TBD
 
 * Introduce ``ZyteAPIResponse`` and ``ZyteAPITextResponse`` which are subclasses
   of ``scrapy.http.Response`` and ``scrapy.http.TextResponse`` respectively.
-  These new response classes hold the raw Zyte API response in its
+  These new response classes hold the raw Zyte Data API response in its
   ``zyte_api`` attribute.
 
 0.1.0 (2022-02-03)
--- a/README.rst
+++ b/README.rst
@@ -94,7 +94,7 @@ key to download a request using Zyte API. Full list of parameters is provided in
         def parse(self, response):
             yield {"URL": response.url, "status": response.status, "HTML": response.body}
 
-            print(response.zyte_api_response)
+            print(response.zyte_api)
             # {
             #     'url': 'https://quotes.toscrape.com/',
             #     'browserHtml': '<html> ... </html>',
@@ -113,7 +113,7 @@ key to download a request using Zyte API. Full list of parameters is provided in
             #     'download_slot': 'quotes.toscrape.com'
             # }
 
-The raw Zyte API Response can be accessed via the ``zyte_api_response`` attribute
+The raw Zyte API Response can be accessed via the ``zyte_api`` attribute
 of the response object. Note that such responses are of ``ZyteAPIResponse`` and
 ``ZyteAPITextResponse`` which are respectively subclasses of ``scrapy.http.Response``
 and ``scrapy.http.TextResponse``. Such classes are needed to hold the raw Zyte API
--- a/README.rst
+++ b/README.rst
@@ -99,7 +99,6 @@ key to download a request using Zyte API. Full list of parameters is provided in
             #     'url': 'https://quotes.toscrape.com/',
             #     'browserHtml': '<html> ... </html>',
             #     'echoData': {'some_value_I_could_track': 123},
-            #     'actions': []
             # }
 
             print(response.request.meta)
--- a/README.rst
+++ b/README.rst
@@ -63,8 +63,12 @@ Here's an example of the things needed inside a Scrapy project's ``settings.py``
 Usage
 -----
 
-To enable every request to be sent through Zyte API, you can set the following
-in the ``settings.py`` file or `any other settings within Scrapy
+To enable a ``scrapy.Request`` to go through Zyte Data API, the ``zyte_api`` key in
+`Request.meta <https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_
+must be present and has dict-like contents.
+
+To set the default parameters for Zyte API enabled requests, you can set the
+following in the ``settings.py`` file or `any other settings within Scrapy
 <https://docs.scrapy.org/en/latest/topics/settings.html#populating-the-settings>`_:
 
 .. code-block:: python
@@ -74,12 +78,12 @@ in the ``settings.py`` file or `any other settings within Scrapy
         "geolocation": "US",
     }
 
-You can see the full list of parameters in the `Zyte API Specification
+You can see the full list of parameters in the `Zyte Data API Specification
 <https://docs.zyte.com/zyte-api/openapi.html#zyte-openapi-spec>`_.
 
-On the other hand, you could also control it on a per-request basis by setting the
-``zyte_api`` key in `Request.meta <https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_.
-When doing so, it will override any parameters set in the 
+Note that the ``ZYTE_API_DEFAULT_PARAMS`` would only work if the ``zyte_api``
+key in `Request.meta <https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_
+is set. When doing so, it will override any parameters set in the 
 ``ZYTE_API_DEFAULT_PARAMS`` setting.
 
 .. code-block:: python
@@ -90,15 +94,19 @@ When doing so, it will override any parameters set in the
     class SampleQuotesSpider(scrapy.Spider):
         name = "sample_quotes"
 
-        def start_requests(self):
+        custom_settings = {
+            "ZYTE_API_DEFAULT_PARAMS": {
+                "geolocation": "US",  # You can set any Geolocation region you want.
+            }
+        }
 
+        def start_requests(self):
             yield scrapy.Request(
                 url="http://books.toscrape.com/",
                 callback=self.parse,
                 meta={
                     "zyte_api": {
                         "browserHtml": True,
-                        "geolocation": "US",  # You can set any Geolocation region you want.
                         "javascript": True,
                         "echoData": {"some_value_I_could_track": 123},
                     }
--- a/README.rst
+++ b/README.rst
@@ -33,8 +33,8 @@ Installation
 
 This package requires Python 3.7+.
 
-How to configure
-----------------
+Configuration
+-------------
 
 Replace the default ``http`` and ``https`` in Scrapy's
 `DOWNLOAD_HANDLERS <https://docs.scrapy.org/en/latest/topics/settings.html#std-setting-DOWNLOAD_HANDLERS>`_
@@ -60,8 +60,8 @@ Here's example of the things needed inside a Scrapy project's ``settings.py`` fi
 
     TWISTED_REACTOR = "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
 
-How to use
-----------
+Usage
+-----
 
 Set the ``zyte_api`` `Request.meta
 <https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_
@@ -70,27 +70,52 @@ key to download a request using Zyte API. Full list of parameters is provided in
 
 .. code-block:: python
 
-   import scrapy
+    import scrapy
 
 
-   class TestSpider(scrapy.Spider):
-       name = "test"
+    class SampleQuotesSpider(scrapy.Spider):
+        name = "sample_quotes"
 
-       def start_requests(self):
+        def start_requests(self):
 
-           yield scrapy.Request(
-               url="http://books.toscrape.com/",
-               callback=self.parse,
-               meta={
-                   "zyte_api": {
-                       "browserHtml": True,
-                       # You can set any GEOLocation region you want.
-                       "geolocation": "US",
-                       "javascript": True,
-                       "echoData": {"something": True},
-                   }
-               },
-           )
+            yield scrapy.Request(
+                url="http://books.toscrape.com/",
+                callback=self.parse,
+                meta={
+                    "zyte_api": {
+                        "browserHtml": True,
+                        "geolocation": "US",  # You can set any Geolocation region you want.
+                        "javascript": True,
+                        "echoData": {"some_value_I_could_track": 123},
+                    }
+                },
+            )
 
-       def parse(self, response):
-           yield {"URL": response.url, "status": response.status, "HTML": response.body}
+        def parse(self, response):
+            yield {"URL": response.url, "status": response.status, "HTML": response.body}
+
+            print(response.zyte_api_response)
+            # {
+            #     'url': 'https://quotes.toscrape.com/',
+            #     'browserHtml': '<html> ... </html>',
+            #     'echoData': {'some_value_I_could_track': 123},
+            #     'actions': []
+            # }
+
+            print(response.request.meta)
+            # {
+            #     'zyte_api': {
+            #         'browserHtml': True,
+            #         'geolocation': 'US',
+            #         'javascript': True,
+            #         'echoData': {'some_value_I_could_track': 123}
+            #     },
+            #     'download_timeout': 180.0,
+            #     'download_slot': 'quotes.toscrape.com'
+            # }
+
+The raw Zyte API Response can be accessed via the ``zyte_api_response`` attribute
+of the response object. Note that such responses are of ``ZyteAPIResponse`` and
+``ZyteAPITextResponse`` which are respectively subclasses of ``scrapy.http.Response``
+and ``scrapy.http.TextResponse``. Such classes are needed to hold the raw Zyte API
+responses.
--- a/README.rst
+++ b/README.rst
@@ -77,9 +77,9 @@ in the ``settings.py`` file or `any other settings within Scrapy
 You can see the full list of parameters in the `Zyte API Specification
 <https://docs.zyte.com/zyte-api/openapi.html#zyte-openapi-spec>`_.
 
-On the other hand, you could also control it on a per request basis by setting the
+On the other hand, you could also control it on a per-request basis by setting the
 ``zyte_api`` key in `Request.meta <https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_.
-When doing so, it will override any parameters that was set in the 
+When doing so, it will override any parameters set in the 
 ``ZYTE_API_DEFAULT_PARAMS`` setting.
 
 .. code-block:: python
@@ -127,8 +127,8 @@ When doing so, it will override any parameters that was set in the
             #     'download_slot': 'quotes.toscrape.com'
             # }
 
-The raw Zyte API Response can be accessed via the ``zyte_api`` attribute
+The raw Zyte Data API response can be accessed via the ``zyte_api`` attribute
 of the response object. Note that such responses are of ``ZyteAPIResponse`` and
-``ZyteAPITextResponse`` which are respectively subclasses of ``scrapy.http.Response``
-and ``scrapy.http.TextResponse``. Such classes are needed to hold the raw Zyte API
+``ZyteAPITextResponse`` types, which are respectively subclasses of ``scrapy.http.Response``
+and ``scrapy.http.TextResponse``. Such classes are needed to hold the raw Zyte Data API
 responses.
--- a/README.rst
+++ b/README.rst
@@ -116,7 +116,7 @@ is set. When doing so, it will override any parameters set in the
         def parse(self, response):
             yield {"URL": response.url, "status": response.status, "HTML": response.body}
 
-            print(response.zyte_api)
+            print(response.raw_api_response)
             # {
             #     'url': 'https://quotes.toscrape.com/',
             #     'browserHtml': '<html> ... </html>',
@@ -135,7 +135,7 @@ is set. When doing so, it will override any parameters set in the
             #     'download_slot': 'quotes.toscrape.com'
             # }
 
-The raw Zyte Data API response can be accessed via the ``zyte_api`` attribute
+The raw Zyte Data API response can be accessed via the ``raw_api_response`` attribute
 of the response object. Note that such responses are of ``ZyteAPIResponse`` and
 ``ZyteAPITextResponse`` types, which are respectively subclasses of ``scrapy.http.Response``
 and ``scrapy.http.TextResponse``. Such classes are needed to hold the raw Zyte Data API
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -15,7 +15,7 @@ from twisted.internet.defer import Deferred, inlineCallbacks
 from zyte_api.aio.client import AsyncClient, create_session
 from zyte_api.aio.errors import RequestError
 
-from .responses import ZyteAPIResponse, ZyteAPITextResponse, process_response
+from .responses import ZyteAPIResponse, ZyteAPITextResponse, _process_response
 
 logger = logging.getLogger(__name__)
 
@@ -88,7 +88,7 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
             raise IgnoreRequest()
 
         self._stats.inc_value("scrapy-zyte-api/request_count")
-        return process_response(api_response, request)
+        return _process_response(api_response, request)
 
     @inlineCallbacks
     def close(self) -> Generator:
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -1,20 +1,21 @@
 import json
 import logging
 import os
-from base64 import b64decode
-from typing import Any, Dict, Generator, List, Optional
+from typing import Any, Dict, Generator, Union
 
 from scrapy import Spider
 from scrapy.core.downloader.handlers.http import HTTPDownloadHandler
 from scrapy.crawler import Crawler
 from scrapy.exceptions import IgnoreRequest, NotConfigured
-from scrapy.http import Request, Response, TextResponse
+from scrapy.http import Request
 from scrapy.settings import Settings
 from scrapy.utils.defer import deferred_from_coro
 from scrapy.utils.reactor import verify_installed_reactor
 from twisted.internet.defer import Deferred, inlineCallbacks
 from zyte_api.aio.client import AsyncClient, create_session
 from zyte_api.aio.errors import RequestError
+
+from .responses import ZyteAPIResponse, ZyteAPITextResponse
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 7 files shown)


scrapy_zyte_api/handler.py
```diff
@@ -81,30 +84,16 @@ async def _download_request(self, request: Request, spider: Spider) -> Response:
...
-            return TextResponse(
-                url=api_response["url"],
-                status=200,
-                body=api_response["browserHtml"].encode(self._encoding),
-                encoding=self._encoding,
-                request=request,
-                flags=["zyte-api"],
-                headers=headers,
+            return ZyteAPITextResponse.from_api_response(
+                api_response, request=request, encoding=self._encoding
```
kmike, line 93: I wonder what is self._encoding for - do you think we can drop it, instead of passing it around?

scrapy_zyte_api/responses.py
```diff
@@ -0,0 +1,52 @@
+from base64 import b64decode
+from typing import Dict, List, Optional
+
+from scrapy import Request
+from scrapy.http import Response, TextResponse
+
+
+class ZyteAPIMixin:
```
kmike, line 8: I wonder if .replace method needs to be implemented as well. Some middlewares might use it to modify the response. I haven't checked which ones though - among built-in middlewares, it's HttpCompressionMiddleware, but there could be others. If I' not mistaken, if we don't override .replace, zyte_api_response data could be lost.

scrapy_zyte_api/handler.py
BurnzZ: Updated in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/d0dc08ddd06b6f21800ccd2dd93af199792c88e6

scrapy_zyte_api/responses.py
```diff
@@ -0,0 +1,52 @@
...
+
+class ZyteAPITextResponse(ZyteAPIMixin, TextResponse):
+    @classmethod
+    def from_api_response(
+        cls, api_response: Dict, *, request: Request = None, encoding: str = "utf-8"
+    ):
+        return cls(
+            url=api_response["url"],
+            status=200,
+            body=api_response["browserHtml"].encode(encoding),
```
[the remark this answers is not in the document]
BurnzZ: Updated in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/d0dc08ddd06b6f21800ccd2dd93af199792c88e6

scrapy_zyte_api/responses.py
BurnzZ: Good catch! Updated it on https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/9695880ef8d20873b06b9f6ba2da335754c54fcf

scrapy_zyte_api/responses.py
```diff
@@ -0,0 +1,52 @@
...
+from scrapy.http import Response, TextResponse
+
+
+class ZyteAPIMixin:
+    def __init__(self, *args, zyte_api_response=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._zyte_api_response = zyte_api_response
+
+    @property
+    def zyte_api_response(self):
```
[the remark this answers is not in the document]
BurnzZ: Updated this on https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/109dbf0f7f4dd6bf54d3cdd2b08504130439c636

README.rst
```diff
@@ -70,27 +70,52 @@ key to download a request using Zyte API. Full list of parameters is provided in
...
+
+        def parse(self, response):
+            yield {"URL": response.url, "status": response.status, "HTML": response.body}
+
+            print(response.zyte_api_response)
+            # {
+            #     'url': 'https://quotes.toscrape.com/',
+            #     'browserHtml': '<html> ... </html>',
+            #     'echoData': {'some_value_I_could_track': 123},
+            #     'actions': []
```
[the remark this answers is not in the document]
BurnzZ: Updated this on https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/109dbf0f7f4dd6bf54d3cdd2b08504130439c636

tests/test_responses.py
```diff
@@ -0,0 +1,122 @@
...
+    "api_response,cls",
+    [
+        (api_response_browser, ZyteAPITextResponse),
+        (api_response_body, ZyteAPIResponse),
+    ],
+)
+def test_response_replace(api_response, cls):
+    orig_response = cls.from_api_response(api_response())
+
+    # The ``zyte_api_response`` should not be replaced.
```
kmike, line 94: Hm, why shouldn't it be possible to replace it? 
If that's a feature which is desirable, but not implemented, we could have an xfail test for this.

BurnzZ: Hi @kmike , this was originally intentional as to prevent the original Zyte API response from being unintentionally replaced. However, I just realized that even other attributes like `url` and `body` could also be replaced. Thus, `zyte_api_response` should have the same expectation like them.

Updated this with an **xfail** in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/84dac7de138f50f6e0727384fafccad6062817a5 as I think the ability to preserve the original Zyte API responses is desirable later on. Kindly let me know otherwise.

Gallaecio:
I was going to suggest changing the class names, because eventually we probably want to allow non-browser responses to be `TextResponse` subclasses (#15).

However, now I am thinking this naming is OK, and what needs to change to address #15 is only that we use `ZyteAPITextResponse` for non-browser responses as well when those are plain-text responses.

I imagine there is a case for users wanting to know if the response content comes from browser or non-browser data, but I don’t think having different classes so that this can be checked with `isinstance` is the right way to go. If anyone requests such a feature, maybe we could add one more property for that (e.g. `response.source == "browserHtml"`).

Gallaecio (commented):
What are you thoughts on this:

- `ZyteAPI*Response` directly maps `browserHtml` if available, else `httpResponseBody`, as proposed. In the future maybe we can allow users to override this behavior through a plugin-specific `zyte_api` meta subkey.
- Whether `ZyteAPIResponse` or `ZyteAPITextResponse` does not depend on whether the body comes from `browserHtml` or `httpResponseBody`, but whether the response is a text response or not.
- `ZyteAPI*Response.api` (rename suggestion for `.zyte_api_response`) is an instance of `ExtractResult` as proposed in https://github.com/zytedata/python-zyte-api/pull/12, which is a `dict` with the API response plus some helpers.
- `ZyteAPI*Response.http` and `ZyteAPI*Response.browser` are regular Scrapy response objects (which class depends on their content and headers) mapping the corresponding content from Zyte Data API, or `None` when not received. They would be lazy-loaded, so that if you do not use `.http` it never gets even base64-decoded.
- `ZyteAPI*Response.replace` works on the top-level response data, but `.http` and `.browser` remain unchanged. (I have not thought too hard about this one, maybe there are scenarios where this is problematic)

BurnzZ:
> ZyteAPI*Response directly maps browserHtml if available, else httpResponseBody, as proposed. In the future maybe we can allow users to override this behavior through a plugin-specific zyte_api meta subkey.

Having `httpResponseBody` and `browserHtml` at the same time isn't allowed by the API as they're mutually exclusive. This is currently the set behavior in the PR wherein `ZyteAPITextResponse` is used when `browserHtml` is present, otherwise `ZyteAPIResponse`.

However, I might be missing something about the notion of "mapping the `browserHtml` when available". Is this related to your last two points?

> Whether ZyteAPIResponse or ZyteAPITextResponse does not depend on whether the body comes from browserHtml or httpResponseBody, but whether the response is a text response or not.

That's a good point. I'll update https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/files#diff-ed75145a6134504c5e20e23c0958e4f4929703abac118e0968f055d51a5fde7bR88 on this one.

> ZyteAPI*Response.api (rename suggestion for .zyte_api_response) is an instance of ExtractResult as proposed in zytedata/python-zyte-api#12, which is a dict with the API response plus some helpers.

Shortening it is indeed better. I just realized that the `*_response` suffix is already redundant (i.e. `response.zyte_api_response`).

I think we can rename it into `zyte_api` instead as it's more clear where the source is coming from. It could be the case that a Scrapy project is using other APIs. Having an explicit `zyte_api` name may alleviate some confusion.

> - ZyteAPI*Response.http and ZyteAPI*Response.browser are regular Scrapy response objects (which class depends on their content and headers) mapping the corresponding content from Zyte Data API, or None when not received. They would be lazy-loaded, so that if you do not use .http it never gets even base64-decoded.
> - ZyteAPI*Response.replace works on the top-level response data, but .http and .browser remain unchanged. (I have not thought too hard about this one, maybe there are scenarios where this is problematic)

I'm afraid I might be missing something here. Would you mind providing some use cases for having `http` and `browser`?

kmike:
> Having httpResponseBody and browserHtml at the same time isn't allowed by the API as they're mutually exclusive. 

This is a limitation which will be addressed in future.

BurnzZ:
> This is a limitation which will be addressed in future.

@kmike I see, thanks for confirming. This answers some of questions from @Gallaecio points above. 👍

Gallaecio:
The fact that `browserHtml` and `httpResponseBody` will be supported simultaneously is what makes https://github.com/zytedata/python-zyte-api/pull/12 such a puzzle, and could affect this as well.

I would not want it to block this PR, but I am worried about keeping backward compatibility if we move on with this PR in a way not compatible with the outcome of https://github.com/zytedata/python-zyte-api/pull/12.

BurnzZ:
> Whether ZyteAPIResponse or ZyteAPITextResponse does not depend on whether the body comes from browserHtml or httpResponseBody, but whether the response is a text response or not.

@kmike @Gallaecio I just realized after working on the above suggestion that we'd have no direct way of knowing if the response is of type `text` _(at least in the current setup)._

For example, `scrapy.Request("https://example.com", callback=self.parse)` without Zyte API would lead to a `response.headers[b'Content-Type']` of value `b'text/html; charset=UTF-8'`. We'd directly know that it's an HTML Text and not an image, file, etc.

But if we have `scrapy.Request("https://example.com", callback=self.parse, meta={"zyte_api": {"httpResponseBody": True}})`, we'd have no access to the headers unless we set `httpResponseHeader=True`.

Thus, by **_default_**, I think we should always set `httpResponseHeader=True` for proper content decoding. Moreover, making the headers always available would make the transition to Zyte API easier for Scrapy projects that relies heavily on accessing the headers.

Any thoughts on this one?

---

On the other hand, as @Gallaecio has pointed out in https://github.com/zytedata/python-zyte-api/pull/12/files#r876821518, we currently have the opportunity to expose some functionalities from the underlying client. 

For example, perhaps we could have `ExtractResult.is_text()` to easily check if its a text response.

kmike:
hey! I think that here we should focus on exposing raw API response, as JSON. Auto-filling httpResponseHeaders also looks out of scope, it's a separate discussion.

@Gallaecio's questions:

> ZyteAPI*Response directly maps browserHtml if available, else httpResponseBody, as proposed. In the future maybe we can allow users to override this behavior through a plugin-specific zyte_api meta subkey.

This sounds good to me. I'm not sure we would even need to override this behavior in future.

> Whether ZyteAPIResponse or ZyteAPITextResponse does not depend on whether the body comes from browserHtml or httpResponseBody, but whether the response is a text response or not.

That's correct. Though browserHtml responses are always text responses.

> ZyteAPI*Response.api (rename suggestion for .zyte_api_response) is an instance of ExtractResult as proposed in https://github.com/zytedata/python-zyte-api/pull/12, which is a dict with the API response plus some helpers.

I'd keep it out of scope here. We can add more helpers later, and discuss it separately. 

> ZyteAPI*Response.http and ZyteAPI*Response.browser are regular Scrapy response objects (which class depends on their content and headers) mapping the corresponding content from Zyte Data API, or None when not received. They would be lazy-loaded, so that if you do not use .http it never gets even base64-decoded.

I think that's fine not to implement it now, and focus on returning the raw data.

> ZyteAPI*Response.replace works on the top-level response data, but .http and .browser remain unchanged. (I have not thought too hard about this one, maybe there are scenarios where this is problematic)

Let's keep this out of scope.

What do you think about implementing the following behavior?

1. For any Zyte API response user should be able to access the raw returned data (decoded from JSON) as `.zyte_api_response` attribute.
2. If there is browserHtml in response, user should be able to run css/xpath selectors, and the response should be a subclass of TextResponse.
3. When only httpResponseBody is in the output (but not httpResponseHeaders), user can't run selectors, and response is not a subclass of TextResponse.
4. When both httpResponseBody and httpResponseHeaders are in response, there should be some kind of content type detection (probably relying on default Scrapy's). If data is text, response should be a subclass of TextResponse, and user should be able to run css/xpath selectors. I guess (4) is what can be tricky to implement; if so, some alternative can be explored for (4).

BurnzZ:
Sounds good to me as well. I've got an idea on how to implement point 4 but testing it with a lot of difference cases is indeed quite tricky. I'll have an update tomorrow after the test cases are sorted out, to see if we're okay with approach.

BurnzZ:
> For any Zyte API response user should be able to access the raw returned data (decoded from JSON) as .zyte_api_response attribute.

Yes, this one is already implemented. Albeit, `.zyte_api_response` has been renamed to `.zyte_api` to prevent redundancies in naming like `response.zyte_api_response`.

> If there is browserHtml in response, user should be able to run css/xpath selectors, and the response type should be a subclass of TextResponse.

This is implemented but I've added some tests in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/910085bfb749cdc9bec0c7092de3d69c280386b5. Thanks for the reminder!

> When only httpResponseBody is in the output (but not httpResponseHeaders), user can't run selectors, and response type is not a subclass of TextResponse.
> When both httpResponseBody and httpResponseHeaders are in response, there should be some kind of content type detection (probably relying on default Scrapy's). If data is text, response type should be a subclass of TextResponse, and user should be able to run css/xpath selectors. I guess (4) is what can be tricky to implement; if so, some alternative can be explored for (4).

Implemented this on https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/e3214d80178169998b06903d71e2b999f25d4eb6

scrapy_zyte_api/responses.py
```diff
@@ -0,0 +1,123 @@
...
+            status=200,
+            body=b64decode(api_response.get("httpResponseBody") or ""),
+            request=request,
+            flags=["zyte-api"],
+            headers=cls._prepare_headers(api_response.get("httpResponseHeaders")),
+            zyte_api=api_response,
+        )
+
+
+def process_response(
```
Gallaecio, line 94: Should we make it private? (`_process_response`)

tests/test_responses.py
Gallaecio: If `zyte_api` is passed to `replace`, then the intention of the coder is to replace it, and I think it should be replaced. If not passed (existing middlewares will not pass it, only new third-party or user ones may), it should stay with the same value.

I wonder if we should turn the property into a regular `zyte_api`-named public attribute, and implement this `replace` method in line with pre-2.6 `TextResponse`:

```python
    def replace(self, *args, **kwargs):
        kwargs.setdefault('zyte_api', self.zyte_api)
        return super().replace(*args, **kwargs)
```

[2.6 introduced Response.attributes](https://github.com/scrapy/scrapy/pull/5218), which I guess we cannot use here since we want support for older Scrapy.

README.rst
```diff
@@ -60,37 +60,75 @@ Here's example of the things needed inside a Scrapy project's ``settings.py`` fi
...
-How to use
-----------
+Usage
+-----
 
-Set the ``zyte_api`` `Request.meta
-<https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_
-key to download a request using Zyte API. Full list of parameters is provided in the
-`Zyte API Specification <https://docs.zyte.com/zyte-api/openapi.html#zyte-openapi-spec>`_.
+To enable every request to be sent through Zyte API, you can set the following
```
kmike, line 66: I'm not sure the setting below makes every request to be sent through Zyte API. If I'm reading the code correctly, it first checks if zyte_api meta key is present and the value is true-ish, and uses ZYTE_API_DEFAULT_PARAMS only in this case. So, parameters are only applied for requests which are already marked explicitly as Zyte API requests.

That's actually the behavior I'm fine with :) It looks useful e.g. to set default geolocation for Zyte API requests made from a spider. On the other hand, making every request going through Zyte API by using this option looks quite problematic; it'd require more thought. For example,

* would this example mean that robots.txt files will be downloaded through Zyte API with browserHtml: True? 
* would it mean that e.g. a POST request won't work properly, while looking like a regular request in the spider code, and without raising any kind of warning? 

It seems some feature to "enable Zyte API transparently" makes sense after we put a bit more effort into "transparent" integration of scrapy Requests with Zyte API direct downloader, it doesn't look as straightforward as setting default parameters.

tests/test_responses.py
BurnzZ: Makes sense. I think we can use the new `Response.attributes` in **2.6** since a lot of the libs we're working on sets it as the minimum version. Updated in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/5b7cf6fded737bb08c0d03b422c50120f920e2ed.

For example, https://github.com/scrapinghub/scrapy-poet/pull/62 needs **2.6** for the non-experimental `asyncio` support.

README.rst
BurnzZ: > I'm not sure the setting below makes every request to be sent through Zyte API. If I'm reading the code correctly, it first checks if zyte_api meta key is present and the value is true-ish, and uses ZYTE_API_DEFAULT_PARAMS only in this case. So, parameters are only applied for requests which are already marked explicitly as Zyte API requests.

That's right. https://github.com/scrapy-plugins/scrapy-zyte-api/pull/13 attempted to not change the behavior fully. It only addressed to prevent the user from repeating the same Zyte API parameters in different parts of the code. In any case, hopefully it should serve as a base point to enable all requests go through Zyte API. :) 

> would this example mean that robots.txt files will be downloaded through Zyte API with browserHtml: True?

That's a good point. This also extends to other things like downloading an image, file, etc.

I think one option is to prevent `browserHtml` from being set in the `ZYTE_API_DEFAULT_PARAMS`. The same is true for `javascript`, `product`, `articles`, etc. I think one way to think of it is having `ZYTE_API_DEFAULT_PARAMS` support only a handful of parameters, like `geolocation` or `httpResponseBody` (though it needs to toggled off when `browserHtml` is turned on), etc.

This could be a bit tricky since Zyte API offers a lot of features that doesn't only cater to being a simply proxy. I think for users to have a seamless experience when using Zyte API, we should have an interface that could cover general Scrapy use. This could mean only downloading the `httpResponseBody` and perhaps `httpResponseHeaders` by default which is available to all `scrapy.Request` unless overridden by the `zyte_api` param in **meta**.

We can open up another PR to explore the different options here.

> would it mean that e.g. a POST request won't work properly, while looking like a regular request in the spider code, and without raising any kind of warning?

I think this could be alleviated somehow by the proposed interface in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/20.

tests/test_responses.py
```diff
@@ -105,22 +105,9 @@ def test_response_replace(api_response, cls):
     new_response = orig_response.replace(url="https://new-example.com")
     assert new_response.url == "https://new-example.com"
```
Gallaecio, line 106: Maybe also add:

```python
assert new_response.zyte_api == api_response()
```

tests/test_responses.py
```diff
@@ -0,0 +1,376 @@
...
+)
+def test_response_replace(api_response, cls):
+    orig_response = cls.from_api_response(api_response())
+
+    # It should still work the same way
+    new_response = orig_response.replace(status=404)
+    assert new_response.status == 404
+
+    new_response = orig_response.replace(url="https://new-example.com")
+    assert new_response.url == "https://new-example.com"
```
Gallaecio, line 106: Maybe also add, after one of these 2 replacements:

```python
assert new_response.zyte_api == api_response()
```__

File: README.rst (hunk 8 above)
Comment (Gallaecio) [reply README.rst:66]: I think @kmike’s main point was that the documentation needs an update, as it is currently quite misleading. The other points are about issues if we decided to make the implementation match the current documentation, which I don’t think we want, at least not at this point.

Comment (BurnzZ) [reply README.rst]: I see, thanks for clarifying! What do you think about this doc update? https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/2adc8a63b984847c693c41cc70d13794213eb237

File: tests/test_responses.py (hunk 10 above)
Comment (BurnzZ) [reply tests/test_responses.py:106]: Good point. Added this in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/32faf3d999622a1b379f355fc515d7803194af27

File: CHANGES.rst [hunk 11]
```diff
@@ -1,6 +1,14 @@
 Changes
 =======
 
+TBD
+---
+
+* Introduce ``ZyteAPIResponse`` and ``ZyteAPITextResponse`` which are subclasses
+  of ``scrapy.http.Response`` and ``scrapy.http.TextResponse`` respectively.
+  These new response classes hold the raw Zyte Data API response in its
+  ``zyte_api`` attribute.
```
kmike, line 10:
```suggestion
  These new response classes hold the raw Zyte Data API response in the
  ``zyte_api`` attribute.
```

Gallaecio: Or maybe “their”.

gatufo: the works for me

scrapy_zyte_api/responses.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if api_response.get("browserHtml"):
+            encoding = _DEFAULT_ENCODING  # Zyte API has "utf-8" by default
+            body = api_response["browserHtml"].encode(encoding)
+        elif api_response.get("httpResponseBody"):
+            body = b64decode(api_response["httpResponseBody"])
+
+        return cls(
+            url=api_response["url"],
+            status=200,
```
gatufo, line 64: Shouldn't this be the site response status?
What if the site returns a 404 with valid content and we want to process that in the spider?

kmike: It should, but Zyte API doesn't have a feature to return the status. It'll be implemented soon, but it's not ready yet.

scrapy_zyte_api/responses.py
```diff
@@ -0,0 +1,123 @@
...
+        # decompress an already decompressed body based on this header.
+        "content-encoding"
+    }
+
+    def __init__(self, *args, zyte_api: Dict = None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._zyte_api = zyte_api
+
+    @property
+    def zyte_api(self) -> Optional[Dict]:
```
kmike, line 25: Naming :) What do you think about calling it with something "raw" in the name? raw_api_response, raw?

I'm still unsure about `response.zyte_api`: response is already a Zyte API response, it's just presented in a different way, mimicking Scrapy response. The difference is not that .zyte_api is a Zyte API resonse, and response is not, the difference is in the format of data; .zyte_api contains the raw, unprocessed version, with all the information available.

tests/test_responses.py
```diff
@@ -0,0 +1,380 @@
...
+
+    # Ensure that the Zyte API response is intact
+    assert new_response.zyte_api == api_response()
+
+    # The Zyte API response must also be fully replaceable
+    new_zyte_api = {"overridden": "value"}
+    new_response = orig_response.replace(zyte_api=new_zyte_api)
+    assert new_response.zyte_api == new_zyte_api
```
kmike, line 114: What should be the behavior when the replaced zyte_api value contains one of the fields used to populate Response attributes? I.e. shouldn't e.g. response.replace(zyte_api={"httpResponseBody": "..."}) change response.body?

I think we should be explicit about it.

scrapy_zyte_api/responses.py
BurnzZ: Updated this in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/cec06773c13f1708af5f80d3533d2202a787c89f

scrapy_zyte_api/responses.py
Gallaecio: `response.raw_api_response` then? Any shorter alternative that sounds nice? `response.raw_data`?

tests/test_responses.py
Gallaecio: It would probably be nice to make it smart that way. Maybe we can extract something out of `_process_response` to handle this well in an overriden `replace`. But we would also need to handle conflicts, e.g. `response.replace(zyte_api={"httpResponseBody": "..."}, body="…")`.

If we do not want to delay this pull request further, maybe we could make it so that `zyte_api` cannot be modified by `replace` (but we still make sure it is not lost either, and handle support to replace it in a follow-up pull request instead.

README.rst
BurnzZ: Hmm a good point. There could be a few cases for this:

1. The user is aware that `ZYTE_API_DEFAULT_PARAMS` is set in the settings and wish to use exactly the same values. Thus, `zyte_api` would be represented as an empty `{}`.
2.  `ZYTE_API_DEFAULT_PARAMS` is not in the settings. The user might've thought that it was set but was actually not. This would result in the requests to be executed without using Zyte API.

We could use `meta={"zyte_api": {}}` to represent it but it kinda feels unnatural as an empty-dict. Perhaps we could also support `meta={"zyte_api": True}` as another alternative. The `{}` is still a valid use-case though since the user might have a dynamic process to determine the contents of meta field. For example:
```python
meta= {"zyte_api": {}}  # default
if does_it_look_like_we_need_javascript():
    meta["zyte_api"].update({"javascript": True})
if how_about_using_fr_region():
    meta["zyte_api"].update({"geolocation": "FR"})
yield scrapy.Request(url, self.callback_func, meta=meta)
```

To summarize, we could use `{}` and `True` to enable Zyte API Requests bearing in mind the need for `ZYTE_API_DEFAULT_PARAMS` be set.

Any thoughts on this?

Gallaecio: +1 to having `{}` and `True` work and behave the same. At some point Zyte Data API may (should? :slightly_smiling_face:) work without parameter other than the URL (e.g. consider `browserHtml` or `httpResponseBody` enabled if no other output is enabled).

scrapy_zyte_api/responses.py
BurnzZ: A good point. I think `response.raw_api` might be good. I believe having the word _"api"_ presents a better context about what the user is trying to access. What do you think?

kmike: I think that's fine for the attribute name to be a bit longer. It seems a concern about `response.raw_api_response` is that "response" is written two times, while "response.raw_api" could already read as "raw api response". But this logic doesn't work for other Response attributes, other `response` attributes are not responses themselves. Because of that, `response.raw_api_response` looks better to me than `response.raw_api`.

kmike: `response.raw_data` also looks fine.

tests/test_responses.py
kmike: This sounds fine to me - preserve .zyte_api on .replace calls, but raise an exception if user calls `response.replace(zyte_api=...)`?

kmike: I'm not even sure what would be a use case for replacing .zyte_api attribute; it could be a feature we won't need for a long time.

BurnzZ: > What should be the behavior when the replaced zyte_api value contains one of the fields used to populate Response attributes? I.e. shouldn't e.g. response.replace(zyte_api={"httpResponseBody": "..."}) change response.body?

Great point! This could be tricky when Zyte API supports both `browserHtml` and `httpResponseBody` at the same time. How would `.replace()` work when both of them are present?

For now, we can be explicit in the test cases wherein performing `.replace()` using **zyte_api** should not result in the related response attribs to be updated. https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/e0865e77e7a627b9a04b2638ca9e3241455d4421

BurnzZ: > but raise an exception if user calls response.replace(zyte_api=...)?

I realized, that I've pushed the commit without refreshing this page to see your new comments. 😅

kmike: Not updating attributes is also not bad, but it seems an exception would be even better.

BurnzZ: Updated it to raise an exception in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/34a427f84fdd6f91d11579c8868d0ff2ccda0c5b

scrapy_zyte_api/responses.py
BurnzZ: > But this logic doesn't work for other Response attributes, other response attributes are not responses themselves. 

You've raised a good point. Renamed it to `response.raw_api_response`. https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/37a4cc705d786a8d091b3066e6426b1f2dc9814c

CHANGES.rst
BurnzZ: Updated in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/37a4cc705d786a8d091b3066e6426b1f2dc9814c

README.rst
kmike: This seems to be the only remaining fix to make, the PR looks good to me otherwise 👍

BurnzZ: Refactored the code to accept `True` and `{}` in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10/commits/f5a9bb074d4f76b8ef294a84edb8c4946919c298.

Gallaecio (approved):
Great work!

kmike:
🎉 thanks @BurnzZ! 🚀

PR #14, opened 2022-05-02
remove 'Content-Encoding' header when returning responses

### LINKED ISSUE #11: Scrapy errors due to decompression attempt based on Content-Encoding
For Zyte API requests with `httpResponseHeader=True`, some websites would return headers like `Content-Encoding: gzip`. 

In the current setup, all of the headers in the `httpResponseHeader` from Zyte API response is used to create the Response in the Download Handler.

Now, when the response containing the `Content-Encoding: gzip` header is processed by the Downloader Middlewares, the `scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware` errors out since it's attempting to decompress the gzipped response.

This is not the case since Zyte API already decompressed the contents.

BurnzZ:
Fixes #11. Built on top of #10.

----

Lots of approach to fix it but it would seem that this is the simplest one. The only downside is that the response instance returned to Scrapy doesn't have the accurate representation of the complete headers since `Content-Encoding` is removed. However, that should be okay since we're preserving it anyways in `response.zyte_api_response.headers`.

### DIFF
--- a/scrapy_zyte_api/responses.py
+++ b/scrapy_zyte_api/responses.py
@@ -8,6 +8,14 @@ _ENCODING = "utf-8"
 
 
 class ZyteAPIMixin:
+
+    REMOVE_HEADERS = {
+        # Zyte API already decompresses the HTTP Response Body. Scrapy's
+        # HttpCompressionMiddleware will error out when it attempts to
+        # decompress an already decompressed body based on this header.
+        "content-encoding"
+    }
+
     def __init__(self, *args, zyte_api_response: Dict = None, **kwargs):
         super().__init__(*args, **kwargs)
         self._zyte_api_response = zyte_api_response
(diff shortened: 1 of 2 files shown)


kmike:
The fix makes sense to me, though argument could be made that it's actually an issue with Zyte API: httpResponseBody is not a raw response, it's returned after content-encoding applied. So, our options:

1. Keep API as-is, remove header here. Pro: simple; I think that's a correct way to tell Scrapy not to decode. Cons: every client may need to do the same, if they support automatic handling of this header. Not sure if that's a big issue.
2. Keep API as-is, remove header here, documnent the API behavior propely in docs.zyte.com.
3. Remove the header in API. Pro: headers would match the content. Cons: headers won't be the same as website returns.
4. Don't decode responses in API, e.g. don't do gunzip. This is the most "pure" way, as it preserves the most information, but it could lead to some confusion on how to work with httpResponseBody, and on increased complexity for clients.

from these options, I like (2) the most. @akshayphilar any preference from you?

akshayphilar:
@kmike agree, the second approach makes sense.

Issue #17, opened 2022-05-13
Map request.headers to requestHeaders or customHttpRequestHeaders


Gallaecio:
I am not sure what to do with incompatible headers, specially with those that are set automatically by Scrapy (e.g. User-Agent). Maybe log a warning if they are user-defined for sure, and silently ignore those set by middlewares?

PR #20, opened 2022-05-17
Map Request.method to httpRequestMethod

### LINKED ISSUE #16: Map Request.method to httpRequestMethod

Gallaecio:
Resolves #16


tests/test_api_requests.py
```diff
@@ -213,6 +215,47 @@ async def test_job_id(self, job_id):
...
+            "DELETE",
+            "GET",
+            "OPTIONS",
+            "PATCH",
+            "POST",
+            "PUT",
+            "TRACE",
+            # Not supported by Zyte Data API at the moment, but processed
+            # nonetheless for forward-compatibility, in case Zyte Data API
+            # starts supporting any of them in the future:
```
BurnzZ, line 230: Thanks for noting these down! 

Currently, whenever we send the `httpRequestMethod` values below, Zyte API returns something like:
```
{"type":"/request/invalid","title":"Bad Request","status":400,"detail":"Cannot construct instance of `com.zyte.gen.model.ExtractRequest$HttpRequestMethodEnum`, problem: Unexpected value 'CONNECT'\n at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 87] (through reference chain: com.zyte.gen.model.ExtractRequest[\"httpRequestMethod\"])"}
```

We'll need to provide a better error message (or at least a warning) when such `httpRequestMethod` values are used. 

On second thought, I think a much better approach would be to block the request from the client-side completely. This also prevents having the user commit an unnecessary and erroneous request.

BurnzZ (changes requested):

tests/test_api_requests.py
Gallaecio: > I think a much better approach would be to block the request from the client-side completely.

I agree as long as the server team confirms that they do not ever plan to support those, i.e. that not supporting them is very explicit, and that they are not willing to add support for them in the future. I imagine that is the case for `CONNECT`, but I am not so sure about `HEAD`.

tests/test_api_requests.py
```diff
@@ -213,6 +215,67 @@ async def test_job_id(self, job_id):
...
+            "DELETE",
+            "GET",
+            "OPTIONS",
+            "PATCH",
+            "POST",
+            "PUT",
+            "TRACE",
+        ],
+    )
+    async def test_supported_method(self, method):
```
BurnzZ, line 230: What do you think about adding the following? as well as only importing `AyncMock` when its `Python3.8+`?
 
```python
@pytest.mark.skipif(
    sys.version_info < (3, 8), reason="Python3.7 has poor support for AsyncMocks"
)
```

In this way, we could pass the tests.

An alternative would be to have workarounds for `AyncMock` like [this one](https://github.com/scrapinghub/web-poet/blob/f2dce80c7c7fc0f3652270843bf01c9b957c4709/tests/test_requests.py#L15-L24)  and [this one](https://github.com/scrapinghub/scrapy-poet/pull/62/files#diff-9e08e8c7769db7b58089fd7659d75ca96df40bb7b5d50299d3a30abd54da9ad6R90-R94).

kmike:
Hey! Tbh, I'm not sure what's the use case for this feature. Does it simplify some code, do you have a "before-after" example?

Gallaecio:
It enables non-GET requests through Zyte Data API. Before we were only able to send GET requests.

kmike:
Weren't you able to do this?
```py
request = Request(
    "http://example.com",
    meta={"zyte_api": {"httpResponseBody": True, "httpRequestMethod": "HEAD"}},
)
```

after:
```py
request = Request(
    "http://example.com",
    method="HEAD",
    meta={"zyte_api": {"httpResponseBody": True}},
)
```

Gallaecio:
@kmike Allowing the change you described in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/20#issuecomment-1131369466 was definitely the goal. The latter feels more Scrapy-like to me, and I wanted to use something like that for the corresponding code example of docs.zyte.com.

kmike:
```py
request = Request(
    "http://example.com",
    method="HEAD",
    meta={"zyte_api": {"httpResponseBody": True}},
)
```

Seems a bit too magical to me; I was thinking about `meta={"zyte_api": ..` as a straightforward, low-level interface, where "zyte_api" content closely corresponds to the parameters sent to the API. "url" is auto-populated there only because Request can't be created without URL, so if it's not populated, there would be duplication.

If we follow this, it could make sense to issue a warning if some non-default Request parameters are used: method, request body. It may be harder to do for headers and cookies, not sure.

On top of this low-level interface we could build some easier to use API for the end users. For example, it could be a custom Request class:

```py
request = ZyteRequest(
    "http://example.com",
    method="HEAD",
)  # {"httpResponseBody": True, "httpResponseHeader": True, httpResponseMethod: request.method, ...?}

request = ZyteBrowserRequest(
    "http://example.com",
    method="HEAD",
)  # {"browserHtml": True, ...?}

# or
request = ZyteRequest(
    "http://example.com",
    method="HEAD",
    use_browser=True,
    actions=[...],
)

```

which can be designed for a common cases.

Gallaecio:
My thinking with https://github.com/scrapy-plugins/scrapy-zyte-api/issues/16 and https://github.com/scrapy-plugins/scrapy-zyte-api/issues/17 was indeed to follow the path of `url`. Even if the reason why `url` is defined that way is an imposition by Scrapy, not allowing other Scrapy-native parameters to behave similarly, specially when there is a 1-to-1 mapping like with `method`, feels a bit inconsistent.

I am OK with keeping things low-level, though. And I don’t see a need for a higher-level API involving custom request classes at the moment.

I do agree that we should log a warning about it, or even log and error and drop the request, when a user specifies a `Request.method` not matching the `zyte_api`-`dict`-specified one. And we can do that for `body` as well. I also have my doubts about headers and cookies, because of the likeliness of them being set by middlewares, possibly not written by the user, including some built-in middlewares; but it might be worth exploring later on, it may be better to over-warn than not to warn at all of what is probably a misuse of a request intended for Zyte Data API.

Gallaecio:
@kmike @BurnzZ I got feedback from @felipeboffnunes about POST requests (where URL is the same, and the difference is in body) being filtered out. Which makes sense since `meta` does not affect request fingerprint.

Since https://github.com/scrapy/scrapy/issues/900 is not addressed yet, I see the following solutions:
- Ask users to use `dont_filter=True` on every request when duplicate URLs could be an issue (this is the only current workaround).
- Create those specialized request classes or request factories we discussed, and make `dont_filter` a part of it.
- Consider mapping properties of `Request` to `zyte_api` parameters as suggested here, if only for proper request fingerprinting.
- Fast-track https://github.com/scrapy/scrapy/issues/900 and create a custom request fingerprinting class that could take into account all `zyte_api` parameters, including e.g. considering 2 requests different if the only difference is that one wants `browserHtml` and the other wants `httpResponseBody`.

kmike:
Providing a custom dupefilter (like in https://github.com/scrapy-plugins/scrapy-splash/blob/master/scrapy_splash/dupefilter.py) is also an option, though it's not composable with other packages which might need custom dupefilters.

That's interesting how all the things are related :) My vote is to fast-track centralized fingerprints.

BurnzZ:
I believe that fast-tracking the centralized fingerprints initiative would be the best long-term solution. However, I'm afraid that it'd be only be available for the new Scrapy version. Say `2.7`?

We might want to consider catering to some Scrapy projects that might have a hard time transitioning to the latest Scrapy version (_for various reasons like their Scrapy project is too large or they don't have resources to do so_). Ultimately, our goal would be a seamless Zyte API integration for Scrapy projects _(using relatively older Scrapy versions)._

With that, I'm inclined to explore the other options that would at least be applicable on the short-term. Here are some of my thoughts:

> Ask users to use dont_filter=True on every request when duplicate URLs could be an issue (this is the only current workaround).

Easiest way to go (I think), since we only need to mention it to our docs. Essentially, it passes the responsibility to the user to do so, instead of `scrapy-zyte-api` doing the work _(or at least Scrapy)_.

However, this is error-prone which leads to a lot of debugging and succeeding fixes by the developer. This would ultimately lead to poor user experience.

> Create those specialized request classes or request factories we discussed, and make dont_filter a part of it.

This removes the strain away from the developer a bit more. 

I think we won't need to set `dont_filter=True` here since if the request factories returns `scrapy.Request` with the body properly set, then it would contribute to a unique request fingerprint. ([reference](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/request.py#L69))

I'm more inclined on this approach than the other option below since `scrapy-zyte-api` has more control over what's happening underneath.

> Consider mapping properties of Request to zyte_api parameters as suggested here, if only for proper request fingerprinting.

kmike:
I think we can address it for users of the most recent Scrapy first, and provide a workaround for previous Scrapy versions separately.

PR #26, opened 2022-05-31
cleanup README

kmike:
* start with an example which doesn't use default parameters
* a few small fixes
* document "zyte_api": True usage pattern

I also think it'd be good to have a version of https://docs.zyte.com/zyte-api/usage/extract.html#decode-html here. I think it's quite unobvious that users should do the following, and not passing `"httpResponseHeaders": True,` is a mistake:

```py
    class SampleQuotesSpider(scrapy.Spider):
        name = "sample_quotes"

        custom_settings = {
            "ZYTE_API_DEFAULT_PARAMS": {
                "httpResponseBody": True,
                "httpResponseHeaders": True,
            }
        }

        def start_requests(self):
            yield scrapy.Request(
                url="http://quotes.toscrape.com/",
                callback=self.parse,
                meta={"zyte_api": True},
            )

        def parse(self, response):
            yield {"URL": response.url, "HTML": response.body}
```

But I'd leave it out of scope for this PR.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,7 +44,7 @@ You also need to set the ``ZYTE_API_KEY``.
 
 Lastly, make sure to `install the asyncio-based Twisted reactor
 <https://docs.scrapy.org/en/latest/topics/asyncio.html#installing-the-asyncio-reactor)>`_
-in the ``settings.py`` file as well:
+in the ``settings.py`` file as well.
 
 Here's an example of the things needed inside a Scrapy project's ``settings.py`` file:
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -135,11 +188,46 @@ is set. When doing so, it will override any parameters set in the
...
             # }
 
-The raw Zyte Data API response can be accessed via the ``raw_api_response`` attribute
-of the response object. Note that such responses are of ``ZyteAPIResponse`` and
-``ZyteAPITextResponse`` types, which are respectively subclasses of ``scrapy.http.Response``
-and ``scrapy.http.TextResponse``. Such classes are needed to hold the raw Zyte Data API
-responses.
+There is a shortcut, in case a request uses the same parameters as
+defined in ``ZYTE_API_DEFAULT_PARAMS`` setting, without any further
+customization - ``zyte_api`` meta key can be set to True:
```
BurnzZ, line 193:
```suggestion
customization - ``zyte_api`` meta key can be set to ``True`` (or even ``{}``):
```

BurnzZ: I believe an explicit mention of the `{}` would be a key tip for users that needs to programmatically set their params like:
```python
meta = {"zyte_api": {}}
if need_geolocation:
    meta["zyte_api"]["geolocation"] = "FR"
yield scrapy.Request(url, self.some_method, meta=meta)
```

BurnzZ (approved):
Thanks @kmike ! 🙌

Gallaecio (approved):
Regardless of my noisy suggestions, I must say I :heart: love the change!

Issue #23, opened 2022-05-27
concurrency handling


kmike:
If I'm not mistaken, currently the download handler uses AsyncClient, which limits concurrency to 15 by default. Probably it would make sense to match it to Scrapy concurrency settings, instead of having this fixed limit. 

When doing so, we need to be careful, and not to forget create_session concurrency options as well.

PR #28, opened 2022-06-07
Adjust n_conn to CONCURRENT_REQUESTS

Gallaecio:
Fixes https://github.com/scrapy-plugins/scrapy-zyte-api/issues/23

I thought about warning if a custom `client` is passed with an `n_conn` different than the one in `CONCURRENT_REQUESTS`, but since that could be done on purpose and providing a setting to silence that warning seemed overkill, I decided against it.

To do:

- [x] Figure out the issue with the 2 failing tests.

### DIFF
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -40,7 +40,10 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
                 )
                 raise NotConfigured
         self._client: AsyncClient = client
-        logger.info(f"Using Zyte API Key: {self._client.api_key[:7]}")
+        logger.info(
+            "Using a Zyte Data API key ending in %r",
+            self._client.api_key[:7]
+        )
         verify_installed_reactor(
             "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
         )
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -41,7 +41,7 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
                 raise NotConfigured
         self._client: AsyncClient = client
         logger.info(
-            "Using a Zyte Data API key ending in %r",
+            "Using a Zyte Data API key starting with %r",
             self._client.api_key[:7]
         )
         verify_installed_reactor(
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -29,7 +29,13 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         if not client:
             try:
                 client = AsyncClient(
-                    api_key=settings.get('ZYTE_API_KEY'),
+                    # To allow users to have a key defined in Scrapy settings
+                    # and in a environment variable, and be able to cause the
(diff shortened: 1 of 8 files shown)


kmike (approved):
The code looks good, but I haven't run it. 

+1 to merge after confirming that the change allows to send more than 15 requests in parallel to Zyte API.

BurnzZ (approved):
LGTM @Gallaecio !

BurnzZ:
Hi @Gallaecio ! Those are some productive improvements! 🎉 I think we can break them down into other PRs as they're out of scope for this PR. Aside from including this commit, +1 on merging this PR to master.

scrapy_zyte_api/handler.py
```diff
@@ -25,28 +26,30 @@ def __init__(
...
+                    n_conn=settings.getint('CONCURRENT_REQUESTS'),
+                )
+            except NoApiKey:
+                logger.warning(
+                    "'ZYTE_API_KEY' must be set in the spider settings or env var "
+                    "in order for ScrapyZyteAPIDownloadHandler to work."
+                )
+                raise NotConfigured
+        self._client: AsyncClient = client
+        logger.info(f"Using Zyte API Key: {self._client.api_key[:7]}")
```
kmike, line 43: Would it make sense to include something like ..., to indicate that's not the full key?

kmike:
Looks awesome @Gallaecio!

For me the main question is ZYTE_API_KEY handling. Let's say a spider sets a certain ZYTE_API_KEY in settings.py. User wants to use env variable. Is it right that previously it was possible by setting the ZYTE_API_KEY to "", and now it requires setting it to None, which is not possible from the command line?

kmike (approved):

BurnzZ:
@kmike , commit https://github.com/scrapy-plugins/scrapy-zyte-api/commit/a255754c1cbd11cdf1cf026eeac026592f7dbb00 seems to be self-documenting enough. LGTM!

Gallaecio:
> Let's say a spider sets a certain ZYTE_API_KEY in settings.py. User wants to use env variable. Is it right that previously it was possible by setting the ZYTE_API_KEY to "", and now it requires setting it to None, which is not possible from the command line?

Do we want to allow API keys to be empty strings?

- If we do, setting the `ZYTE_API_KEY` setting to an empty string should make `""` the API key. Maybe we can support a special string in settings that would translate to the key not being set, e.g. `NONE` or `UNSET`.
- If we do not, then this may be worth handling [upstream](https://github.com/zytedata/python-zyte-api/blob/main/zyte_api/apikey.py#L12-L20).

kmike:
> Do we want to allow API keys to be empty strings?

Ideally, yes. But I think that's fine if scrapy-zyte-api doesn't allow them.

> If we do, setting the ZYTE_API_KEY setting to an empty string should make "" the API key. Maybe we can support a special string in settings that would translate to the key not being set, e.g. NONE or UNSET.

I'm not sure how to do it for scrapy-zyte-api, if only a string value can be set from cmdline (is it?). In this case, "UNSET" string doesn't look any better than "". If we're not talking about command line, None already works.

EIther way, it looks like a very minor edge case; I'm not against merging any of the implementations.

Testing failure looks like a large issue though, as it could mean tests become unstable; could you please check it?

kmike:
Nice, thanks @Gallaecio!

Issue #22, opened 2022-05-26
Define a minimum Scrapy version


Gallaecio:
We should indicate a minimum Scrapy version in the README.

I also wonder if Scrapy should be a dependency, `setup.py`-wise, so that the right version of Scrapy (or, more likely, the other way around) can be enforced during installation. [scrapy-zyte-smartproxy does it](/https://github.com/scrapy-plugins/scrapy-zyte-smartproxy/blob/master/setup.py#L38).

BurnzZ:
I think we could perhaps set **2.6**, as the minimum as briefly discussed in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10#discussion_r883248928.

+1 on enforcing it in the `install_requires` param.

PR #29, opened 2022-06-13
update dep requirement: scrapy>=2.6.0

BurnzZ:
Some users encounter problems like the one below when using the old version of Scrapy:
```
ERROR: <twisted.python.failure.Failure scrapy.exceptions.NotSupported: Unsupported URL scheme 'https': type object 'TextResponse' has no attribute 'attributes'>
```

It stems from using the new `Response.attributes` of Scrapy 2.6.

I'm proposing to require users to have a minimum of `scrapy>=2.6.0` to be able to use it.

Fixes https://github.com/scrapy-plugins/scrapy-zyte-api/issues/22

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ setuptools.setup(
     author_email="info@zyte.com",
     url="https://github.com/scrapy-plugins/scrapy-zyte-api",
     packages=["scrapy_zyte_api"],
-    install_requires=["zyte-api>=0.1.2"],
+    install_requires=["zyte-api>=0.1.2", "scrapy>=2.6.0"],
     classifiers=[
         "Development Status :: 4 - Beta",
         "Intended Audience :: Developers",
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changes
 =======
 
+TBD
+---
+
+* Update minimum requirement: ``scrapy>=2.6.0``.
+
 
 0.2.0 (2022-05-31)
 ------------------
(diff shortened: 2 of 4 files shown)


Gallaecio (approved):

setup.py
```diff
@@ -10,7 +10,7 @@
     author_email="info@zyte.com",
     url="https://github.com/scrapy-plugins/scrapy-zyte-api",
     packages=["scrapy_zyte_api"],
-    install_requires=["zyte-api>=0.1.2"],
+    install_requires=["zyte-api>=0.1.2", "scrapy>=2.6.0"],
```
kmike, line 13: Given https://github.com/pypa/pip/pull/4500 is there, it could be ok to have scrapy >= 2.6.0 in install_requires.
On the other hand, let's say I'm using scrapy 2.4, and want to use Zyte API. I install scrapy-zyte-api, but suddenly Scrapy gets upgraded, probably breaking things in the project.

What do you think about failing at runtime (or at import time?), instead of using install_requires? For the scenario above it seems like a better behavior: you get a clear message that Scrapy 2.6+ is required, and you can decide on how you're going to approach it.

kmike: > Also, in the scenario where users have an older Scrapy and there is a version of scrapy-zyte-api that supports that version, things would just work™ with a requirements.txt with Scrapy<x.y.z and scrapy-zyte-api. With the run-time approach, users would need to manually find out which version of scrapy-zyte-api to install.

That's nice, I haven't considered it. It looks like my packaging knowledge is pretty outdated :) Thanks @Gallaecio!

kmike:
Thanks @BurnzZ!

PR #30, opened 2022-07-06
Allow overriding the API request retry policy

Gallaecio:


kmike (approved):
Nice! 👍 for adding a real-world example.

README.rst
```diff
@@ -232,3 +232,48 @@ customization - the ``zyte_api`` meta key can be set to ``True`` or ``{}``:
             #     'download_slot': 'quotes.toscrape.com'
             # }
 
+Customizing the retry policy
+----------------------------
+
+API requests are retried automatically using the default retry policy of
+`python-zyte-api`_.
```
BurnzZ, line 239: Optional: perhaps it might be convenient if we could add a link to https://github.com/zytedata/python-zyte-api/blob/main/zyte_api/aio/retry.py#L61-L131 so users could immediately know what to subclass and its existing behavior.

scrapy_zyte_api/handler.py
```diff
@@ -95,9 +93,12 @@ async def _download_request(
         api_data = {**{"url": request.url}, **api_params}
         if self._job_id is not None:
             api_data["jobId"] = self._job_id
+        retrying = request.meta.get("zyte_api_retry_policy") or self._retry_policy
         try:
             api_response = await self._client.request_raw(
-                api_data, session=self._session
+                api_data,
+                session=self._session,
+                retrying=retrying,
```
BurnzZ, line 101: minor: I was thinking if we should set `handle_retries=True` to ensure that `retrying` is always used in the client: https://github.com/zytedata/python-zyte-api/blob/main/zyte_api/aio/client.py#L111-L112

BurnzZ (approved):
Aside from a few minor things, LGTM @Gallaecio ! Good job!

scrapy_zyte_api/handler.py
Gallaecio: Given that is the default value, and this is a public API (i.e. the default value should stay like that for backward compatibility), I don’t think it is necessary, but I do not mind adding it either (explicit better than implicit?).

Gallaecio: Oh, I forgot we allowed custom clients. I actually see your point now, if the configured retry policy is anything other than `None` maybe we should pass `handle_retries=True` as well.

That said, custom clients with `handle_retries=False` as default could also be smart about `retrying`, i.e. if `retrying` is not `None`, ignore `handle_retries` and work as if `handle_retries=True`.

All-in-all, I think the best solution going forward may be to deprecate `handle_retries` and instead provide a `NO_RETRIES` retry policy that can be passed to `retrying` to disable retries. And maybe rename `retrying` to `retry_policy`.

I will drop the idea as suggested, but I am completely OK with forcing `handle_retries=True` if a custom retrying policy is defined in settings or request metadata, so let me know if you change your mind.

README.rst
```diff
@@ -232,3 +232,50 @@ customization - the ``zyte_api`` meta key can be set to ``True`` or ``{}``:
...
+
+        def stop(self, retry_state: RetryCallState) -> bool:
+            if is_http_521(retry_state.outcome.exception()):
+                return self.temporary_download_error_stop(retry_state)
+            return super().stop(retry_state)
+
+    ZYTE_API_RETRY_POLICY = CustomRetryFactory().build()
+
+.. _python-zyte-api: https://github.com/zytedata/python-zyte-api
+.. _RetryFactory: https://github.com/zytedata/python-zyte-api/blob/5b38f47c50552a5c0fcdc7cd4084a1b6bf2428d6/zyte_api/aio/retry.py#L61-L131
```
BurnzZ, line 280:
```suggestion
.. _RetryFactory: https://github.com/zytedata/python-zyte-api/blob/main/zyte_api/aio/retry.py#L61-L131
```

Gallaecio: I did that on purpose so that the lines do not break. Maybe we could just link to the file itself?

BurnzZ: Yeah, let's link the files instead. My concern with having the commit hash in the link is that it might be pointing to an old code. Having it in `master` ensures the freshest version.

scrapy_zyte_api/handler.py
```diff
@@ -47,18 +47,16 @@ def __init__(
...
         self._client: AsyncClient = client
         logger.info(
-            "Using a Zyte Data API key starting with %r",
-            self._client.api_key[:7]
+            "Using a Zyte Data API key starting with %r", self._client.api_key[:7]
         )
         verify_installed_reactor(
             "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
         )
         self._stats = crawler.stats
```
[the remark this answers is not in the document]
kmike: Great feedback, thanks @starrify!

PR #33, opened 2022-07-28
Expose python-zyte-api stats to Scrapy

### LINKED ISSUE #31: Expose python-zyte-api stats as Scrapy stats
See https://github.com/scrapy-plugins/scrapy-zyte-api/pull/30#discussion_r919151176

Gallaecio:
Resolves #31

I am slightly worried about the impact that copying stats (specially means) can have on CPU-bound spiders.

I first tried to make it so that stats would be copied every N seconds with reactor.callLater, and finally when the download handler closes. To my surprise, that did not work when doing a manual test. As far as I can tell, the `close` method of a download handler is never called, the downloader seems to only call `download_request`.

So I went for copying after every request download.

Do you think this is OK? Should we consider skipping the mean stats from the copy?

To do:
- [x] Agree on an approach
- [x] Add tests

### DIFF
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -92,14 +92,22 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
             '429',
             'attempts',
             'errors',
-            'extracted_queries',
             'fatal_errors',
-            'input_queries',
-            'results',
+            'processed',
+            'success',
         ):
             self._stats.set_value(
                 f"{prefix}/{stat}",
                 getattr(self._client.agg_stats, f"n_{stat}"),
+            )
+        for stat in (
+            'error_ratio',
+            'success_ratio',
+            'throttle_ratio',
+        ):
+            self._stats.set_value(
+                f"{prefix}/{stat}",
+                getattr(self._client.agg_stats, stat)(),
             )
         for source, target in (
             ('connect', 'connection'),
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -120,7 +120,6 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
                     value,
                 )
 
-
     async def _download_request(
         self, api_params: dict, request: Request, spider: Spider
     ) -> Optional[Union[ZyteAPITextResponse, ZyteAPIResponse]]:
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -146,8 +146,8 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
                 f"Got an error when processing Zyte API request ({request.url}): {er}"
             )
             raise IgnoreRequest()
-
-        self._update_stats()
+        finally:
+            self._update_stats()
 
         return _process_response(api_response, request)
 
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -86,6 +86,41 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
             raise IgnoreRequest()
         return api_params
 
+    def _update_stats(self):
+        prefix = "scrapy-zyte-api"
+        for stat in (
+            '429',
+            'attempts',
+            'errors',
+            'extracted_queries',
+            'fatal_errors',
+            'input_queries',
+            'results',
+        ):
+            self._stats.set_value(
+                f"{prefix}/{stat}",
+                getattr(self._client.agg_stats, f"n_{stat}"),
+            )
+        for source, target in (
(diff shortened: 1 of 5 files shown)


scrapy_zyte_api/handler.py
```diff
@@ -111,8 +146,9 @@ async def _download_request(
                 f"Got an error when processing Zyte API request ({request.url}): {er}"
             )
             raise IgnoreRequest()
+        finally:
+            self._update_stats()
 
-        self._stats.inc_value("scrapy-zyte-api/request_count")
```
Gallaecio, line 115: Should we keep this stat?

Gallaecio:
Hmm… It just crossed my mind that I could use a close signal handler to ensure that stats are gathered on exit. Shall I move to the callLater approach with an interval between gatherings? (e.g. 60 seconds, configurable through a setting)

BurnzZ:
Another approach could be something like https://github.com/scrapinghub/scrapinghub-entrypoint-scrapy/blob/master/sh_scrapy/stats.py using `twisted.internet.task.LoopingCall`. Though I think updating the stats for every `_download_request()` call is better since it's always in sync.

scrapy_zyte_api/handler.py
BurnzZ: The new equivalent of this would be `scrapy-zyte-api/input_queries` right?

The only concern here would be some users relying on checking this stat. Though I think that could easily be addressed by reflecting it clearly in the Changelog.

kmike:
> I am slightly worried about the impact that copying stats (specially means) can have on CPU-bound spiders.

Don't worry about computing means and other stats; we use running averages, provided by https://github.com/grantjenks/python-runstats/package, it should be O(1) operation, just getting a value which is already computed.

kmike:
If you're worried about performance, we could benchmark it, but overall the approach looks good to me. So, +1 to proceed with it.

kmike:
The default limit on Zyte API key is currently 2RPS, so stats updates shouldn't be much more frequent than once every 0.5s; overhead should be negligible.

kmike (approved):
Lookg good @Gallaecio! 

I'm not sure I like all the stats names which we have in python-zyte-api, but that's a separate discussion. ~~I don't have a good proposal now, so it'd be better not to block this feature on it. Names are not awful, they may be good enough :)~~

Sorry, let me think a bit about it. For example, it seems that e.g. n_extracted_queries is always the same as n_results.

kmike:
What do you think about doing it after a python-zyte-api release with https://github.com/zytedata/python-zyte-api/pull/27 is ready (if the PR makes sense)? Minimum python-zyte-api version would need to be updated in setup.py.

scrapy_zyte_api/handler.py
```diff
@@ -86,6 +86,40 @@ def _prepare_api_params(self, request: Request) -> Optional[dict]:
...
+        for counter in (
+            'api_error_types',
+            'exception_types',
+            'status_codes',
+        ):
+            for key, value in getattr(self._client.agg_stats, counter).items():
+                self._stats.set_value(
+                    f"{prefix}/{counter}/{key}",
+                    value,
+                )
```
kmike, line 121: What do you think about exposing success_ratio, throttle_ration, and maybe error_ratio as well? 

For error_ratio my concern is that the name might suggest it's just 1-success_ratio, but that's a very different statistic.

Gallaecio: Sounds good. I would probably fall on that trap myself, but I don’t think it is a big deal. We can always add `1-success_ratio` as failure_ratio, in case its better contrast with success_ration helps understand that error_ratio is something else.

kmike: a great idea about failure ratio!

kmike: would it be ok to add failure ratio for scrapy-zyte-api stats only?

Gallaecio: I am also OK with not adding it at all. As I said, I don‘t think the potential misinterpretation of error_ratio is a big deal, and we can probably also approach this with documentation _if_ people even complain.

kmike: either approach is fine with me

setup.py
```diff
@@ -10,7 +10,7 @@
     author_email="info@zyte.com",
     url="https://github.com/scrapy-plugins/scrapy-zyte-api",
     packages=["scrapy_zyte_api"],
-    install_requires=["zyte-api>=0.1.2", "scrapy>=2.6.0"],
+    install_requires=["zyte-api>=0.3.0", "scrapy>=2.6.0"],
```
Gallaecio, line 13: Note to self: bump2version is going to screw this up as well.

kmike:
Thanks @Gallaecio!

setup.py
kmike: of course, I tried to make a release and forgot about it :)

Issue #12, opened 2022-04-29
Ability to set the default Zyte API Parameters for every Request


BurnzZ:
Currently, the parameters are passed on a per-request basis.

Users should be able to:

- enable Zyte API for all request
- set the default parameters for every request

BurnzZ:
The PR in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/13 only partially addresses the points (2) of the issue as it doesn't introduce the mechanism of how to properly enable every `scrapy.Request` to go through Zyte API without having the need to set the `zyte_api` param in Request.meta.

There are also things we need to figure out like in the https://github.com/scrapy-plugins/scrapy-zyte-api/pull/10#discussion_r882801793 thread.

PR #40, opened 2022-08-12
Add ZYTE_API_ENABLED

Gallaecio:
Resolves https://github.com/scrapy-plugins/scrapy-zyte-api/issues/12

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,6 +60,9 @@ Here's an example of the things needed inside a Scrapy project's ``settings.py``
 
     TWISTED_REACTOR = "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
 
+The ``ZYTE_API_ENABLED`` setting, which is ``True`` by default, can be set to
+``False`` to disable this plugin.
+
 Usage
 -----
 
@@ -213,9 +216,6 @@ go through Zyte Data API unless the ``zyte_api`` key in Request.meta_ is set to
 Zyte Data API requests need parameters. If you set the ``zyte_api`` key in
 Request.meta_ or the ``ZYTE_API_ON_ALL_REQUESTS`` setting to ``True``, you must
 also :ref:`set default parameters <default-params>`.
-
-Set the ``ZYTE_API_ENABLED`` setting to ``False`` to disable this plugin.
-``ZYTE_API_ENABLED`` is ``True`` by default.
 
 
 Customizing the retry policy
--- a/README.rst
+++ b/README.rst
@@ -116,8 +116,12 @@ and ``scrapy.http.TextResponse``.
 If multiple requests target the same URL with different Zyte Data API
 parameters, pass ``dont_filter=True`` to ``Request``.
 
+
+.. _default-params:
+
 Setting default parameters
 --------------------------
+
 Often the same configuration needs to be used for all Zyte API requests.
 For example, all requests may need to set the same geolocation, or
 the spider only uses ``browserHtml`` requests.
@@ -133,13 +137,7 @@ following in the ``settings.py`` file or `any other settings within Scrapy
         "geolocation": "US",
     }
 
-
-``ZYTE_API_DEFAULT_PARAMS`` works if the ``zyte_api``
-key in `Request.meta <https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_
-is set, i.e. having ``ZYTE_API_DEFAULT_PARAMS`` doesn't make all requests
-to go through Zyte Data API. Parameters in ``ZYTE_API_DEFAULT_PARAMS`` are 
-merged with parameters set via the ``zyte_api`` meta key, with the values in 
-meta taking priority.
+For example:
 
 .. code-block:: python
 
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -191,49 +189,36 @@ meta taking priority.
...
+Controlling which requests go through Zyte Data API
+---------------------------------------------------
 
-        custom_settings = {
-            "ZYTE_API_DEFAULT_PARAMS": {
-                "browserHtml": True,
-            }
-        }
+The ``ZYTE_API_ENABLED`` setting can be used to control whether all, none, or
+some requests go through Zyte Data API. It supports the following values:
```
kmike, line 205: I think it might be better to have a separate option for enabling all requests to go through Zyte API, because of the following reasons:

1. Usually ..._ENABLED options just turn on/off the middleware.
2. It may also be tricky to set this setting to None from the command line, if I recall correctly. 
3. We shoud consider future Zyte API features, with data extraction use cases: if user only wants the extraction features, they'd need to set ZYTE_API_ENABLED to None, which is a bit strange.
4. Instead of relying on the user to specify correct request parameters to use Zyte API as a downloader, it'd be nice to set them on our side. This may involve not only passing static default parameters, but also taking request method, body and headers in account.

README.rst
```diff
@@ -191,49 +189,36 @@ meta taking priority.
...
 
-        def start_requests(self):
-            yield scrapy.Request(
-                url="http://quotes.toscrape.com/",
-                callback=self.parse,
-                meta={"zyte_api": True},
-            )
+-   ``None`` (default): only requests where the ``zyte_api`` key in
+    Request.meta_ is set to ``True`` or set to a dictionary go through Zyte
+    Data API.
```
BurnzZ, line 209: I'm not too sure about this part, since at first glance, the `ZYTE_API_ENABLED` setting seems to only need `True` or `False` with some boolean default set to it. 

What do you think about having the following?
- `ZYTE_API_ENABLED=False` turn off the plugin completely.
- `ZYTE_API_ENABLED=True` _(default)_ turn it on unless the `zyte_api` key in `Request.meta` is set explicitly set to `False`.

and then a separate setting to control the behavior of the defaults.

README.rst
Gallaecio: 3 is not fully clear to me, but I don’t think it matters at this point.

Shall we deal with 4 on a follow-up pull request, or here? I would love to work on it, but I also would like to discuss the API a bit before working on an implementation.

Gallaecio:
Feel free to suggest alternatives to `ZYTE_API_ALL` (`ZYTE_API_ALL_REQUESTS`? `ZYTE_API_FORCE`? `ZYTE_API_USE_BY_DEFAULT`?)

README.rst
```diff
@@ -191,49 +189,32 @@ meta taking priority.
...
-            yield scrapy.Request(
-                url="http://quotes.toscrape.com/",
-                callback=self.parse,
-                meta={"zyte_api": True},
-            )
+.. _Request.meta: https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta
 
-        def parse(self, response):
-            yield {"URL": response.url, "HTML": response.body}
+Set the ``ZYTE_API_ALL`` setting to ``True`` to make all requests go through
```
BurnzZ, line 209: I think `ZYTE_API_ON_ALL_REQUESTS: bool` sounds more clear.

Let's also mention the default value here.

scrapy_zyte_api/handler.py
```diff
@@ -66,10 +70,16 @@ def download_request(self, request: Request, spider: Spider) -> Deferred:
...
 
     def _prepare_api_params(self, request: Request) -> Optional[dict]:
-        meta_params = request.meta.get("zyte_api")
+        meta_params = request.meta.get("zyte_api", self._all)
+        if meta_params is False:
+            return None
         if not meta_params and meta_params != {}:
+            warn(
+                f"Setting the zyte_api request metadata key to "
+                f"{meta_params!r} is deprecated. Use False instead.",
```
BurnzZ, line 79: If the `meta_params` value is set to `False`, this message would be triggered.

Since we have this in the README:
> Set the `ZYTE_API_ALL` setting to `True` to make all requests go through Zyte Data API unless the `zyte_api` key in [Request.meta](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta) is set to `False`.

then I think this warning message might be confusing.

Do we mean somethin else here?

Gallaecio: > If the `meta_params` value is set to `False`, this message would be triggered.

It shouldn’t due to the early return above.

BurnzZ: Ahhh gotcha, we're checking `None` here. 👍

Gallaecio: I did not limit it to `meta_params is None` because the original code was not checking that either, so it is technically possible for someone to use `"zyte_api": 0` or similar. I don’t think anyone ever did, but just in case.

tests/test_api_requests.py
```diff
@@ -134,6 +136,126 @@ async def test_http_response_headers_request(meta: Dict[str, Dict[str, Any]], mo
...
+        ({"zyte_api": {}}, {"ZYTE_API_ON_ALL_REQUESTS": False}, {}, False),
+        ({"zyte_api": True}, {"ZYTE_API_ON_ALL_REQUESTS": False}, {}, False),
+        ({"zyte_api": False}, {"ZYTE_API_ON_ALL_REQUESTS": False}, {}, False),
+        (
+            {},
+            {
+                "ZYTE_API_DEFAULT_PARAMS": {"browserHtml": True, "geolocation": "CA"},
+                "ZYTE_API_ON_ALL_REQUESTS": False,
+            },
+            {"browserHtml": True, "geolocation": "CA"},
```
BurnzZ, line 151: Shouldn't this be an empty `{}` instead?

tests/test_api_requests.py
```diff
@@ -137,6 +139,124 @@ async def test_http_response_headers_request(
             {"browserHtml": True, "geolocation": "US", "javascript": True},
             True,
         ),
+        # ZYTE_API_ON_ALL_REQUESTS=False
```
BurnzZ, line 142: Since this is the default value, it's quite redundant with the test cases above that omits it.

My concern here is that the test cases is a bit harder to maintain because of it since changing/adding something would mean that 2 test cases are involved.

tests/test_api_requests.py
```diff
@@ -137,6 +139,124 @@ async def test_http_response_headers_request(
...
+            {"zyte_api": False},
+            {
+                "ZYTE_API_DEFAULT_PARAMS": {"browserHtml": True, "geolocation": "CA"},
+                "ZYTE_API_ON_ALL_REQUESTS": True,
+            },
+            {},
+            False,
+        ),
+        (
+            {"zyte_api": None},
```
BurnzZ, line 225: In the current README docs, it's described as:

> Set the ``ZYTE_API_ON_ALL_REQUESTS`` setting to ``True`` to make all requests
go through Zyte Data API unless the ``zyte_api`` key in Request.meta_ is set to
``False``. ``ZYTE_API_ON_ALL_REQUESTS`` is ``False`` by default.

I'm concerned that the docs isn't portraying the correct message since a value of `None` doesn't go through ZDA while the test case below with `{}` goes through.

Should we disable the ZDA calls _**strictly**_ if the value is `False` and not _false-like_? 

Alternatively, we could also mention that both `False` and `None` values are supported in the README.

Gallaecio: `None` works, just like `0`, but these changes include deprecating them, and that is why they are not even mentioned in the documentation.

Users should use `False` or a `dict`. Anything else is deprecated. For backward compatibility `None` and `0` will work as before, but they should not be used in any new code.

tests/test_api_requests.py
Gallaecio: Switched to a generator to keep the test cases without code repetition.

tests/test_api_requests.py
BurnzZ: Gotcha. I totally forgot about https://github.com/scrapy-plugins/scrapy-zyte-api/pull/40/files#diff-ed75145a6134504c5e20e23c0958e4f4929703abac118e0968f055d51a5fde7bR80-R84

BurnzZ (approved):
LGTM @Gallaecio !

Gallaecio:
Closing in favor of https://github.com/scrapy-plugins/scrapy-zyte-api/pull/41

Issue #19, opened 2022-05-16
Enable httpResponseHeaders by default if httpResponseBody is enabled


Gallaecio:
> for Scrapy it's probably an anti-pattern to request httpResponseBody, but not httpResponseHeaders, as the encoding detection may not work properly in this case. — @kmike 

We could probably set it to True by default, but still allow users to disable it on purpose if they wish (I don’t see why, but it should be trivial to allow them).

PR #41, opened 2022-08-12
Support transparent mapping of regular requests to Zyte Data API requests

Gallaecio:
Built on top of https://github.com/scrapy-plugins/scrapy-zyte-api/pull/40, this pull request aims to address point 4 from https://github.com/scrapy-plugins/scrapy-zyte-api/pull/40#discussion_r944261278.

Please, ignore the actual implementation at the time being, and instead have a look at the test scenarios introduced here. I think we should discuss changes, additions and removals around those before moving forward. Questions to discuss include:
- Are we OK with making the feature opt-out, or should it be opt-in?
- Better suggestions for new settings? Are the two future-proofing-related settings worth adding?
- Any additional considerations to make the plugin more future-proof? Does any of the approaches go too far?
- Are we OK with the approach to warning, and the scenarios causing a warning? Should we be more lenient with headers about warnings? Is it OK to warn against request metadata usage for all keys that can be defined through request attributes?

To do:
- [ ] Agree on specifications based on test scenarios.
- [ ] Refactor the implementation.
- [ ] Provide documentation.

Resolves https://github.com/scrapy-plugins/scrapy-zyte-api/issues/16, resolves https://github.com/scrapy-plugins/scrapy-zyte-api/issues/17, resolves https://github.com/scrapy-plugins/scrapy-zyte-api/issues/19.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -116,8 +116,12 @@ and ``scrapy.http.TextResponse``.
 If multiple requests target the same URL with different Zyte Data API
 parameters, pass ``dont_filter=True`` to ``Request``.
 
+
+.. _default-params:
+
 Setting default parameters
 --------------------------
+
 Often the same configuration needs to be used for all Zyte API requests.
 For example, all requests may need to set the same geolocation, or
 the spider only uses ``browserHtml`` requests.
@@ -133,13 +137,7 @@ following in the ``settings.py`` file or `any other settings within Scrapy
         "geolocation": "US",
     }
 
-
-``ZYTE_API_DEFAULT_PARAMS`` works if the ``zyte_api``
-key in `Request.meta <https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_
-is set, i.e. having ``ZYTE_API_DEFAULT_PARAMS`` doesn't make all requests
-to go through Zyte Data API. Parameters in ``ZYTE_API_DEFAULT_PARAMS`` are 
-merged with parameters set via the ``zyte_api`` meta key, with the values in 
-meta taking priority.
+For example:
 
 .. code-block:: python
 
@@ -191,49 +189,36 @@ meta taking priority.
             #     'download_slot': 'quotes.toscrape.com'
             # }
 
-There is a shortcut, in case a request uses the same parameters as
-defined in the ``ZYTE_API_DEFAULT_PARAMS`` setting, without any further
-customization - the ``zyte_api`` meta key can be set to ``True`` or ``{}``:
+``ZYTE_API_DEFAULT_PARAMS`` does not make requests automatically go through
+Zyte Data API. See :ref:`enabled`.
 
-.. code-block:: python
-
-    import scrapy
+Parameters in ``ZYTE_API_DEFAULT_PARAMS`` are merged with parameters set via
+the ``zyte_api`` meta key, with the values in meta taking priority.
 
 
-    class SampleQuotesSpider(scrapy.Spider):
-        name = "sample_quotes"
+.. _enabled:
 
-        custom_settings = {
-            "ZYTE_API_DEFAULT_PARAMS": {
-                "browserHtml": True,
-            }
-        }
+Controlling which requests go through Zyte Data API
+---------------------------------------------------
 
-        def start_requests(self):
-            yield scrapy.Request(
-                url="http://quotes.toscrape.com/",
-                callback=self.parse,
-                meta={"zyte_api": True},
-            )
+The ``ZYTE_API_ENABLED`` setting can be used to control whether all, none, or
+some requests go through Zyte Data API. It supports the following values:
 
-        def parse(self, response):
-            yield {"URL": response.url, "HTML": response.body}
+-   ``None`` (default): only requests where the ``zyte_api`` key in
+    Request.meta_ is set to ``True`` or set to a dictionary go through Zyte
+    Data API.
 
-            print(response.raw_api_response)
-            # {
-            #     'url': 'https://quotes.toscrape.com/',
-            #     'statusCode': 200,
-            #     'browserHtml': '<html> ... </html>',
-            # }
+    .. _Request.meta: https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta
 
-            print(response.request.meta)
-            # {
-            #     'zyte_api': {
-            #         'browserHtml': True,
-            #     },
-            #     'download_timeout': 180.0,
-            #     'download_slot': 'quotes.toscrape.com'
-            # }
+-   ``True``: all requests go through Zyte Data API, unless the ``zyte_api``
+    key in Request.meta_ is set to ``False``.
+
+-   ``False``: disables this plugin.
+
+Zyte Data API requests need parameters. You must either set those parameters in
+the ``zyte_api`` metadata key of every request or :ref:`set default parameters
+<default-params>`.
+
 
 Customizing the retry policy
 ----------------------------
--- a/README.rst
+++ b/README.rst
@@ -206,14 +206,16 @@ By default, only requests where the ``zyte_api`` key in Request.meta_ is set to
 
 .. _Request.meta: https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta
 
-Set the ``ZYTE_API_ALL`` setting to ``True`` to make all requests go through
-Zyte Data API unless the ``zyte_api`` key in Request.meta_ is set to ``False``.
+Set the ``ZYTE_API_ON_ALL_REQUESTS`` setting to ``True`` to make all requests
+go through Zyte Data API unless the ``zyte_api`` key in Request.meta_ is set to
+``False``. ``ZYTE_API_ON_ALL_REQUESTS`` is ``False`` by default.
 
 Zyte Data API requests need parameters. If you set the ``zyte_api`` key in
-Request.meta_ or the ``ZYTE_API_ALL`` setting to ``True``, you must also
-:ref:`set default parameters <default-params>`.
+Request.meta_ or the ``ZYTE_API_ON_ALL_REQUESTS`` setting to ``True``, you must
+also :ref:`set default parameters <default-params>`.
 
 Set the ``ZYTE_API_ENABLED`` setting to ``False`` to disable this plugin.
+``ZYTE_API_ENABLED`` is ``True`` by default.
 
 
 Customizing the retry policy
--- a/README.rst
+++ b/README.rst
@@ -201,23 +201,19 @@ the ``zyte_api`` meta key, with the values in meta taking priority.
 Controlling which requests go through Zyte Data API
 ---------------------------------------------------
 
-The ``ZYTE_API_ENABLED`` setting can be used to control whether all, none, or
-some requests go through Zyte Data API. It supports the following values:
+By default, only requests where the ``zyte_api`` key in Request.meta_ is set to
+``True`` or set to a dictionary go through Zyte Data API.
 
--   ``None`` (default): only requests where the ``zyte_api`` key in
-    Request.meta_ is set to ``True`` or set to a dictionary go through Zyte
-    Data API.
+.. _Request.meta: https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta
 
-    .. _Request.meta: https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta
+Set the ``ZYTE_API_ALL`` setting to ``True`` to make all requests go through
+Zyte Data API unless the ``zyte_api`` key in Request.meta_ is set to ``False``.
 
--   ``True``: all requests go through Zyte Data API, unless the ``zyte_api``
-    key in Request.meta_ is set to ``False``.
+Zyte Data API requests need parameters. If you set the ``zyte_api`` key in
+Request.meta_ or the ``ZYTE_API_ALL`` setting to ``True``, you must also
+:ref:`set default parameters <default-params>`.
 
--   ``False``: disables this plugin.
-
-Zyte Data API requests need parameters. You must either set those parameters in
-the ``zyte_api`` metadata key of every request or :ref:`set default parameters
-<default-params>`.
+Set the ``ZYTE_API_ENABLED`` setting to ``False`` to disable this plugin.
 
 
 Customizing the retry policy
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -1,5 +1,6 @@
 import logging
 from typing import Any, Dict, Generator, Optional, Union
+from warnings import warn
 
 from scrapy import Spider
 from scrapy.core.downloader.handlers.http import HTTPDownloadHandler
@@ -25,6 +26,11 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         self, settings: Settings, crawler: Crawler, client: AsyncClient = None
     ):
         super().__init__(settings=settings, crawler=crawler)
+        enabled = settings.get("ZYTE_API_ENABLED")
+        if enabled is False:
+            raise NotConfigured
+        self._enabled_by_default = enabled or False
+
         if not client:
             try:
                 client = AsyncClient(
@@ -66,10 +72,16 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         return super().download_request(request, spider)
 
     def _prepare_api_params(self, request: Request) -> Optional[dict]:
-        meta_params = request.meta.get("zyte_api")
-        if not meta_params and meta_params != {}:
+        meta_params = request.meta.get("zyte_api", self._enabled_by_default)
+        if meta_params is False:
             return None
-
+        if not meta_params and meta_params != {}:
+            warn(
+                f"Setting the zyte_api request metadata key to "
+                f"{meta_params!r} is deprecated. Use False instead.",
+                DeprecationWarning,
+            )
+            return None
         if meta_params is True:
             meta_params = {}
 
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -1,4 +1,5 @@
 import logging
+from base64 import b64decode, b64encode
 from typing import Any, Dict, Generator, Optional, Union
 from warnings import warn
 
@@ -8,6 +9,10 @@ from scrapy.crawler import Crawler
 from scrapy.exceptions import IgnoreRequest, NotConfigured
 from scrapy.http import Request
 from scrapy.settings import Settings
+from scrapy.settings.default_settings import (
+    DEFAULT_REQUEST_HEADERS,
+    USER_AGENT as DEFAULT_USER_AGENT
+)
 from scrapy.utils.defer import deferred_from_coro
 from scrapy.utils.reactor import verify_installed_reactor
 from twisted.internet.defer import Deferred, inlineCallbacks
@@ -60,6 +65,21 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         self._session = create_session(connection_pool_size=self._client.n_conn)
         self._retry_policy = settings.get("ZYTE_API_RETRY_POLICY")
         self._on_all_requests = settings.getbool("ZYTE_API_ON_ALL_REQUESTS")
+        self._automap = settings.getbool("ZYTE_API_AUTOMAP", True)
+        self._unsupported_headers = {
+            header.strip().lower().encode() for header in settings.getlist(
+                "ZYTE_API_UNSUPPORTED_HEADERS",
+                ["Cookie", "User-Agent"],
+            )
+        }
+        browser_headers = settings.getdict(
+            "ZYTE_API_BROWSER_HEADERS",
+            {"Referer": "referer"},
+        )
+        self._browser_headers = {
+            k.strip().lower().encode(): v
+            for k, v in browser_headers.items()
+        }
 
     def download_request(self, request: Request, spider: Spider) -> Deferred:
         api_params = self._prepare_api_params(request)
@@ -93,6 +113,145 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
                 f"instead ({request.url})."
             )
             raise IgnoreRequest()
+
+        if not self._automap:
+            return api_params
+
+        if not any(
+            api_params.get(k)
+            for k in ("httpResponseBody", "browserHtml", "screenshot")
+        ):
+            api_params.setdefault("httpResponseBody", True)
+        response_body = api_params.get("httpResponseBody")
+
+        if any(api_params.get(k) for k in ("httpResponseBody", "browserHtml")):
+            if api_params.get("httpResponseHeaders") is True:
+                logger.warning(
+                    "You do not need to set httpResponseHeaders to True if "
+                    "you httpResponseBody or browserHtml to True. Note that "
+                    "httpResponseBody is set to True automatically if neither "
+                    "browserHtml nor screenshot are set to True."
+                )
+            api_params.setdefault("httpResponseHeaders", True)
+
+        method = api_params.get("httpRequestMethod")
+        if method:
+            logger.warning(
+                f"Request {request} uses the Zyte Data API httpRequestMethod "
+                f"parameter. Use Request.method instead."
+            )
+            if method != request.method:
+                logger.warning(
+                    f"The HTTP method of request {request} ({request.method}) "
+                    f"does not match the Zyte Data API httpRequestMethod "
+                    f"parameter ({method})."
+                )
+        elif request.method != "GET":
+            if response_body:
+                api_params["httpRequestMethod"] = request.method
+            else:
+                logger.warning(
+                    f"The HTTP method of request {request} ({request.method}) "
+                    f"is being ignored. The httpRequestMethod parameter of "
+                    f"Zyte Data API can only be set when the httpResponseBody "
+                    f"parameter is True."
+                )
+
+        if response_body:
+            headers = api_params.get("customHttpRequestHeaders")
+            if headers is not None:
+                logger.warning(
+                    f"Request {request} defines the Zyte Data API "
+                    f"customHttpRequestHeaders parameter. Use Request.headers "
+                    f"instead."
+                )
+            elif request.headers:
+                headers = []
+                for k, v in request.headers.items():
+                    if not v:
+                        continue
+                    v = b','.join(v).decode()
+                    lowercase_k = k.strip().lower()
+                    if lowercase_k in self._unsupported_headers:
+                        if (
+                            lowercase_k != b'user-agent'
+                            or v != DEFAULT_USER_AGENT
+                        ):
+                            logger.warning(
+                                f"Request {request} defines header {k}, which "
+                                f"cannot be mapped into the Zyte Data API "
+                                f"customHttpRequestHeaders parameter."
+                            )
+                        continue
+                    k = k.decode()
+                    headers.append({"name": k, "value": v})
+                if headers:
+                    api_params["customHttpRequestHeaders"] = headers
+        if (
+            not response_body
+            or any(api_params.get(k) for k in ("browserHtml", "screenshot"))
+        ):
+            headers = api_params.get("requestHeaders")
+            if headers is not None:
+                logger.warning(
+                    f"Request {request} defines the Zyte Data API "
+                    f"requestHeaders parameter. Use Request.headers instead."
+                )
+            elif request.headers:
+                request_headers = {}
+                for k, v in request.headers.items():
+                    if not v:
+                        continue
+                    v = b','.join(v).decode()
+                    lowercase_k = k.strip().lower()
+                    key = self._browser_headers.get(lowercase_k)
+                    if key is not None:
+                        request_headers[key] = v
+                    elif not (
+                        (
+                            lowercase_k == b'accept'
+                            and v == DEFAULT_REQUEST_HEADERS['Accept']
+                        ) or (
+                            lowercase_k == b'accept-language'
+                            and v == DEFAULT_REQUEST_HEADERS['Accept-Language']
+                        ) or (
+                            lowercase_k == b'user-agent'
+                            and v == DEFAULT_USER_AGENT
+                        )
+                    ):
+                        logger.warning(
+                            f"Request {request} defines header {k}, which "
+                            f"cannot be mapped into the Zyte Data API "
+                            f"requestHeaders parameter."
+                        )
+                if request_headers:
+                    api_params["requestHeaders"] = request_headers
+
+        body = api_params.get("httpRequestBody")
+        if body:
+            logger.warning(
+                f"Request {request} uses the Zyte Data API httpRequestBody "
+                f"parameter. Use Request.body instead."
+            )
+            decoded_body = b64decode(body)
+            if decoded_body != request.body:
+                logger.warning(
+                    f"The body of request {request} ({request.body!r}) "
+                    f"does not match the Zyte Data API httpRequestBody "
+                    f"parameter ({body!r}; decoded: {decoded_body!r})."
+                )
+        elif request.body != b"":
+            if response_body:
+                base64_body = b64encode(request.body).decode()
+                api_params["httpRequestBody"] = base64_body
+            else:
+                logger.warning(
+                    f"The body of request {request} ({request.body!r}) "
+                    f"is being ignored. The httpRequestBody parameter of "
+                    f"Zyte Data API can only be set when the httpResponseBody "
+                    f"parameter is True."
+                )
+
         return api_params
 
     def _update_stats(self):
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -26,11 +26,8 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         self, settings: Settings, crawler: Crawler, client: AsyncClient = None
     ):
         super().__init__(settings=settings, crawler=crawler)
-        enabled = settings.get("ZYTE_API_ENABLED")
-        if enabled is False:
+        if not settings.getbool("ZYTE_API_ENABLED", True):
             raise NotConfigured
-        self._enabled_by_default = enabled or False
-
         if not client:
             try:
                 client = AsyncClient(
@@ -62,6 +59,7 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         self._zyte_api_default_params = settings.getdict("ZYTE_API_DEFAULT_PARAMS")
         self._session = create_session(connection_pool_size=self._client.n_conn)
         self._retry_policy = settings.get("ZYTE_API_RETRY_POLICY")
+        self._all = settings.getbool("ZYTE_API_ALL")
 
     def download_request(self, request: Request, spider: Spider) -> Deferred:
         api_params = self._prepare_api_params(request)
@@ -72,7 +70,7 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         return super().download_request(request, spider)
 
     def _prepare_api_params(self, request: Request) -> Optional[dict]:
-        meta_params = request.meta.get("zyte_api", self._enabled_by_default)
+        meta_params = request.meta.get("zyte_api", self._all)
         if meta_params is False:
             return None
         if not meta_params and meta_params != {}:
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -59,7 +59,7 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         self._zyte_api_default_params = settings.getdict("ZYTE_API_DEFAULT_PARAMS")
         self._session = create_session(connection_pool_size=self._client.n_conn)
         self._retry_policy = settings.get("ZYTE_API_RETRY_POLICY")
-        self._all = settings.getbool("ZYTE_API_ALL")
+        self._on_all_requests = settings.getbool("ZYTE_API_ON_ALL_REQUESTS")
 
     def download_request(self, request: Request, spider: Spider) -> Deferred:
         api_params = self._prepare_api_params(request)
@@ -70,7 +70,7 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         return super().download_request(request, spider)
 
     def _prepare_api_params(self, request: Request) -> Optional[dict]:
-        meta_params = request.meta.get("zyte_api", self._all)
+        meta_params = request.meta.get("zyte_api", self._on_all_requests)
         if meta_params is False:
             return None
         if not meta_params and meta_params != {}:
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -2,6 +2,7 @@ from contextlib import asynccontextmanager, contextmanager
 from os import environ
 from typing import Optional
 
+from scrapy.exceptions import NotConfigured
 from scrapy.utils.misc import create_instance
 from scrapy.utils.test import get_crawler
 from twisted.internet.asyncioreactor import AsyncioSelectorReactor
@@ -30,15 +31,19 @@ async def make_handler(settings: dict, api_url: Optional[str] = None):
     if api_url is not None:
         settings["ZYTE_API_URL"] = api_url
     crawler = get_crawler(settings_dict=settings)
-    handler = create_instance(
-        ScrapyZyteAPIDownloadHandler,
-        settings=None,
-        crawler=crawler,
-    )
+    try:
+        handler = create_instance(
+            ScrapyZyteAPIDownloadHandler,
+            settings=None,
+            crawler=crawler,
+        )
+    except NotConfigured:
+        handler = None
     try:
         yield handler
     finally:
-        await handler._close()  # NOQA
+        if handler is not None:
+            await handler._close()  # NOQA
 
 
 @contextmanager
--- a/tests/test_api_requests.py
+++ b/tests/test_api_requests.py
@@ -1,6 +1,6 @@
 import sys
 from asyncio import iscoroutine
-from typing import Any, Dict
+from typing import Any, Dict, List, Literal, Union
 from unittest import mock
 
 import pytest
@@ -9,6 +9,10 @@ from pytest_twisted import ensureDeferred
 from scrapy import Request, Spider
 from scrapy.exceptions import IgnoreRequest, NotSupported
 from scrapy.http import Response, TextResponse
+from scrapy.settings.default_settings import (
+    DEFAULT_REQUEST_HEADERS,
+    USER_AGENT as DEFAULT_USER_AGENT,
+)
 from scrapy.utils.defer import deferred_from_coro
 from scrapy.utils.test import get_crawler
 from twisted.internet.defer import Deferred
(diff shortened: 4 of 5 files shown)


tests/test_api_requests.py
```diff
@@ -310,3 +506,742 @@ async def parse(self, response):
...
+            },
+            {},
+            {
+                "httpResponseBody": True,
+                "httpResponseHeaders": True,
+                "httpRequestMethod": "PATCH",
+            },
+            [
+                "Use Request.method instead",
+                "does not match the Zyte Data API httpRequestMethod parameter",
```
BurnzZ, line 731: To make the warning message more clear, we can mention which value is actually used.

The same is true for the header warnings in the test cases below.

tests/test_api_requests.py
```diff
@@ -310,3 +506,742 @@ async def parse(self, response):
...
+                "httpResponseHeaders": True,
+            },
+            ["cannot be mapped"],
+        ),
+        # You may update the ZYTE_API_UNSUPPORTED_HEADERS setting to remove
+        # headers that the customHttpRequestHeaders parameter starts supporting
+        # in the future.
+        (
+            {
+                "headers": {"User-Agent": ""},
```
BurnzZ, line 1082: Should we add the `Cookie` header here to test out `ZYTE_API_UNSUPPORTED_HEADERS`?

BurnzZ (commented):
> Are we OK with making the feature opt-out, or should it be opt-in

I'm leaning towards having it opt-out (enabled by default) since it seems more natural to write the requests this way. +1 on the current setup.

> Better suggestions for new settings? Are the two future-proofing-related settings worth adding?

Could you clarify the expected behaviors for `ZYTE_API_UNSUPPORTED_HEADERS` and `ZYTE_API_BROWSER_HEADERS`? I'm not entirely sure if I understood how they work.

> Any additional considerations to make the plugin more future-proof? Does any of the approaches go too far?

I think the `ZYTE_API_AUTOMAP` captures most of the cases that we need. Great work! 

> Are we OK with the approach to warning, and the scenarios causing a warning? Should we be more lenient with headers about warnings? Is it OK to warn against request metadata usage for all keys that can be defined through request attributes?

The route of issuing a warning sounds great. Though I think we need to tweak it a bit to make it clear for the users what is actually passed _(i.e. the values taking precedence)_ when warnings are used.

kmike:
hey! I was thinking about being able to do the following 3 things in a spider, at the same time:

1) Make all "normal" requests go through Zyte API instead of regular Scrapy downloader, in a way which is similar to using a proxy for everything. 
2) It'd be nice to be able to control some parameters for these "normal" requests per-request.
3) Make some requests to use the exact Zyte API options provided, without any magic. There is going to be an UI to debug these parameters, finiding those which are working, and copy-pasting them.

Use case: extract data from some pages, or to use some browser action, while downloading everything else as usual.

So, it seems it's neither "opt-in" nor "opt-out", it's almost like two separate features, which exist in parallel. Does it make sense?

Gallaecio:
1 would be the default behavior with the proposed implementation, and could be disabled through a setting. 2 would also work with the proposed implementation.

For 3, what about defining a `zyte_api_automap` request meta key that can be set to `False` per request?

Gallaecio:
> Could you clarify the expected behaviors for `ZYTE_API_UNSUPPORTED_HEADERS` and `ZYTE_API_BROWSER_HEADERS`? I'm not entirely sure if I understood how they work.

First off: I am more than open to remove them altogether.

The idea is for these parameters to allow flexibility to support some future Zyte API changes without needing to upgrade to a newer scrapy-zyte-api version.

If tomorrow Zyte API starts allowing to set the `Cookie` or `User-Agent` headers, you can use the `ZYTE_API_UNSUPPORTED_HEADERS`setting to make it so that scrapy-zyte-api stops excluding them from the mapping. So, if you set the `User-Agent` header, and set `ZYTE_API_UNSUPPORTED_HEADERS=['Cookies']`, the `User-Agent` header is included in `customHttpRequestHeaders`.

Similar for `ZYTE_API_BROWSER_HEADERS`. If in the future Zyte API allows an additional header, like `Accept-Language`, you can set `ZYTE_API_BROWSER_HEADERS={"Accept-Language": "acceptLanguage", "Referer": "referer"}` so that the header gets mapped into `requestHeaders` as `acceptLanguage`.

scrapy_zyte_api/handler.py
```diff
@@ -86,6 +114,149 @@ def _prepare_api_params(self, request: Request) -> Optional[dict]:
                 f"instead. ({request})."
             )
             raise
+
+        if request.meta.get("zyte_api_automap", self._automap):
```
kmike, line 118: So, the current behavior (writing it down for myself mostly):

Configuration 1: all default options

```py
yield Request(url)  # regular request

yield Request(url, meta={"zyte_api": true})  # Zyte API request with body, headers, method, etc.

# Zyte API request with body, headers, method, etc., but changes the geolocation
yield Request(url, meta={"zyte_api": {"geolocation": "uk"}}) 

# browserHtml requests; some options are not supported (e.g. request method)
yield Request(url, meta={"zyte_api": {"browserHtml": True}}) 

# exact Zyte API request, no magic
yield Request(url, meta={"zyte_api": {"browserHtml": True}, "zyte_api_automap": False}) 

```

Configuration 2: "proxy mode". ZYTE_API_ON_ALL_REQUESTS=True

```py
yield Request(url)  # Zyte API request with body, headers, method, etc.

# Zyte API request with body, headers, method, etc., but with UK geolocation
yield Request(url, meta={"zyte_api": {"geolocation": "uk"}}) 

# browserHtml request; a different mapping of headers (?), less supported Request options
yield Request(url, meta={"zyte_api": {"browserHtml": True}}) 

yield Request(url, meta={"zyte_api": False})  # regular request

# exact Zyte API request, no magic
yield Request(url, meta={"zyte_api": {"browserHtml": True}, "zyte_api_automap": False}) 
```

scrapy_zyte_api/handler.py
```diff
@@ -86,6 +114,149 @@ def _prepare_api_params(self, request: Request) -> Optional[dict]:
...
+            self._map_params(api_params, request)
+
+        return api_params
+
+    def _map_headers(self, api_params, request):
```
kmike, line 123: Could you try converting _map_headers and _map_params to some utility methods, which work without a need to create a handler? It might simplify testing.

scrapy_zyte_api/handler.py
```diff
@@ -86,6 +114,149 @@ def _prepare_api_params(self, request: Request) -> Optional[dict]:
...
+                            f"Request {request} defines header {k}, which "
+                            f"cannot be mapped into the Zyte Data API "
+                            f"requestHeaders parameter."
+                        )
+                if request_headers:
+                    api_params["requestHeaders"] = request_headers
+
+    def _map_params(self, api_params, request):
+        if not any(
+            api_params.get(k) for k in ("httpResponseBody", "browserHtml", "screenshot")
```
kmike, line 193: this is the piece I'm least certain about, from the API point of view. Let's say automatic mapping is enabled. User sends the following request:

```py
yield Request(url, meta={"zyte_api": {"screenshot": True}})
```

Would you expect to get a screenshot in addition to the usual response, or would you expect to get screenshot only? If that's screenshot only, should response body be set to the screenshot content?

What happens if we add data extraction later, e.g. 

```py
yield Request(url, meta={"zyte_api": {"product": True}})
```

If we follow the same approach, would it mean that response is going to be empty, and only product data would be returned?

If we're in a "Zyte API is my downloader" mindset, it's arguably expected that `response` in a callback still works as usual, but there is an additional data in `response` attributes (currently in `response.raw_api_response`, but we can improve on that).

This may lead to some inefficient code though, when users only want a screenshot or data, but we automatically request httpResponseBody as well.

Sorry for rambling, I'm still processing it :) 

I have a vague idea that making automatic mode completely separate from manual mode could help - i.e. instead of using `zyte_api` meta key for both modes, use separate meta keys for them (e.g. `zyte_api` and `zyte_api_raw`), as semantics of the parameters there are different. But it needs more thought.

Gallaecio: > ```python
> yield Request(url, meta={"zyte_api": {"screenshot": True}})
> ```
> 
> Would you expect to get a screenshot in addition to the usual response, or would you expect to get screenshot only?

I think we can choose what we think will be the most commonly desired scenario, and make that the default, as long as we document it. Similar to what we do with response headers.

For example, let’s say we believe that most commonly when users set `"screenshot": True` they will also want `browserHtml`. We can set `browserHtml` to `True` in those scenarios, but still allow to override that setting `browserHtml` to `False`.

I think that behavior would be acceptable as long as we document it.

However, since to get `browserHtml` users need, with the current implementation, to explicitly request it, I think in this specific case it may be best for `screenshot` to get only the screenshot, and not browser HTML. And, as you mention later on:

> This may lead to some inefficient code though, when users only want a screenshot or data, but we automatically request httpResponseBody as well.

---

> If that's screenshot only, should response body be set to the screenshot content?
>
> What happens if we add data extraction later, e.g.
> 
> ```python
> yield Request(url, meta={"zyte_api": {"product": True}})
> ```
> 
> If we follow the same approach, would it mean that response is going to be empty, and only product data would be returned?

I think this question is valid regardless of whether or not automatic mapping is enabled. So I don’t think we need an answer as part of this pull request.

I have no strong opinion on what the right behavior should be.

I think it would be weird for response to be anything other than an HTTP response, so if you send a request for e.g. only an automatically-extracted product, I don’t think it needs to become the response, I think an empty response with `.zyte_api_raw['product']` or directly `.product` may be best. But it could become the response as JSON, I guess.

For a future where you can combine browserHtml with httpResponseBody and we want users to be able to control which one becomes the response (which I am not sure is needed, honestly), or in case we want to allow screenshot to become the response, we could have a default implementation like “browser, else http, else screenshot” and expose a request meta key and setting to change that per-request or globally (e.g. `zyte_api_body_source="screenshot"`).

> I have a vague idea that making automatic mode completely separate from manual mode could help - i.e. instead of using `zyte_api` meta key for both modes, use separate meta keys for them (e.g. `zyte_api` and `zyte_api_raw`), as semantics of these parameters are different. But it needs more thought.

It may be a good approach. It is shorter than having a separate `automap` meta key. I do wonder how well it would play with other settings like `DEFAULT_PARAMS`.

Shall I extend tests (and implementation) to explore the scenarios with this approach?

Gallaecio: By the way, if the future-proofing settings are not too much, we could consider adding one that determines what keywords are considered content-defining, so that `httpResponseBody ` is not added if they are added. Thought that may be going too far, and it opens up other questions: Do we need to have separate settings to define which content keywords are compatible with browserHtml and which are not? Do we need to do the same for httpResponseBody? Do we need to allow configuring content keywords incompatible with both? Do we need to allow defining additional content keywords that are incompatible with each other?

Gallaecio:
Tests refactored to use the new _get_api_params instead of instantiating a mock server every time. They are also split into groups, but otherwise are basically the same as before.

However, I also made the following changes as I was working on that refactoring:
- Backward incompatible: if the meta object is truthy but not a mapping, `ValueError` is now raised (before it would raise `TypeError` or `ValueError` depending on the value type)
- Added additional warnings when defining in meta parameters with no effect (removing them would cause the same API parameters to be used in the end due to the automatic mapping logic).
- Automated mapping now removes parameters if set to their server-side default value, minimizing parameters sent to the server.
- `None` as a meta parameter value unsets a parameter set through `ZYTE_API_DEFAULT_PARAMS`. If we fear `null` may become a non-default, valid value for an API parameter, we could go for a special `UNSET` object. I added this at first to allow silencing warnings caused by combinations of default params and meta params that made sense separately, but I later decided to solve that in a way that does not require explicitly unsetting params, and now I am not sure if we should keep this feature or remove it. I did not come up with the alternative either; the scenarios I could think of are covered by `test_default_params_automap` and trigger a warning with the current implementation although I think they should not.
- Explicitly defined `httpResponseBody=False`, `browserHtml=False` and `screenshot=False` in contexts where they are not necessary triggers a warning.

The implementation is growing horrible, but I think we can make it clean once we figure out what behavior we want API-wise.

tests/test_api_requests.py
```diff
@@ -325,3 +228,1149 @@ async def parse(self, response):
...
+        # The body is copied into httpRequestBody, base64-encoded.
+        (
+            {
+                "body": "a",
+            },
+            {},
+            {
+                "httpResponseBody": True,
+                "httpResponseHeaders": True,
+                "httpRequestBody": "YQ==",
```
BurnzZ, line 400: Should we also ensure that the `httpRequestMethod` is not `GET` when the body is set?

tests/test_api_requests.py
```diff
@@ -325,3 +228,1149 @@ async def parse(self, response):
...
+                "requestHeaders": {"referer": "a"},
+            },
+            [],
+        ),
+        (
+            {"Referer": "a"},
+            {"browserHtml": True, "httpResponseBody": True},
+            {
+                "browserHtml": True,
+                "customHttpRequestHeaders": [
```
BurnzZ, line 1074: To clarify, wouldn't using `customHttpRequestHeaders` alongside `browserHtml=True` issue a warning?

Since the [doc](https://docs.zyte.com/zyte-api/openapi.html#operation/extract) mentions:
> `customHttpRequestHeaders`: This feature cannot be used with browserHtml, only with httpResponseBody

setup.py
```diff
@@ -10,7 +10,11 @@
     author_email="info@zyte.com",
     url="https://github.com/scrapy-plugins/scrapy-zyte-api",
     packages=["scrapy_zyte_api"],
-    install_requires=["zyte-api>=0.3.0", "scrapy>=2.6.0"],
+    install_requires=[
+        "scrapy>=2.6.0",
+        "typing-extensions>=3.10",
```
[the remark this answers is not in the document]
Gallaecio: That was my initial approach, then https://github.com/python/mypy/issues/8520 hit me :slightly_frowning_face:

scrapy_zyte_api/handler.py
kmike: It looks like there are 2-3 significantly different use cases:

1. Use Zyte API as a replacement for Scrapy downloader, essentially as a proxy.
2. Everything else, specialized use cases: screenshots, data extraction.
3. A combined use case, where you want both direct download results and some extra data, e.g. screenshot or extracted data.

browserHtml is somewhere in between (1) and (2), probably closer to the "proxy" side.

The best API for these use cases could be different. 

For example, for (1) you'd like all requests to be passed through Zyte API implicitly. You'd also like to support various scrapy.Request parameters, and behave in a transparent way, in general. Sometimes you may need to e.g. replace httpResponseBody with browserHtml, set actions, and maybe tweak geolocation. To access the data, you'd want to use response.body/text, or run response.css/xpath selectors.

For (2) you want only sepecific requests to be handled, it should be opt-in. Likely you'd want to always specify request parameters, such as data type to extract, or screenshot size. It's not clear how useful is it to support Request parameters other than URL; it may be nice in some cases, but there's a lot of caveats. It doesn't make much sense to put extracted data to response body, and I think the same applies for screenshots.

For (3) you'd like scrapy.Request and response parameters to be mapped transparently, like in (1). In addition to that, you'd need to specify request parameters, and have this mode opt-in, like in (2).

What do you think about the following proposal?

1. Add an option like `ZYTE_API_DOWNLOAD=True`. This enables Mode (1) - automatic mapping of scrapy.Request, etc. For the user it looks like everything works as before, but Zyte API is now used to download data.
2. Add a request.meta key (something like `zyte_api_extra`), which allows to add more things to the API request: set geolocation, request screenshot. When something is present in this meta key, it's an extra, the default httpResponseBody is still present. This enables configuration for Mode (1), and it also enables Mode (3).
3. When `zyte_api` meta key is present, disable Mode 1, if it's enabled and use parameters in`zyte_api` as-is. This is essentially Mode (2). It should be backwards compatible to what we have in the main branch.
4. When `browserHtml` is present in `zyte_api_extra`, switch to "browser mode" - don't request httpResponseBody, use browserHtml as response body. I.e. returning both httpResponseBody and browserHtml in the same request is not supported with Mode (1); users can do it using Mode (2), with `zyte_api` meta keys.

Gallaecio: > Add an option like `ZYTE_API_DOWNLOAD=True`

Naming-wise, don’t you prefer `ZYTE_API_AUTOMAP` for this, or something in those lines (e.g. TRANSPARENT_MODE, PROXY_MODE)? To me `DOWNLOAD` feels generic enough that it could be confusing, since you can “download” content with `DOWNLOAD=False`, i.e. someone that still does not know what it does would wonder how `ENABLED` and `DOWNLOAD` are different.

> browserHtml handling in the proposal above is a bit weak.

I am not sure what you mean here, it sounds OK to me.

---

Besides naming, I am OK with this approach of split meta keys, I’ll go for it.

tests/test_api_requests.py
```diff
@@ -26,9 +31,19 @@
...
+        {"zyte_api": {"httpResponseBody": True}},
+        {"zyte_api": {"httpResponseBody": True, "geolocation": "US"}},
+        {
+            "zyte_api": {
+                "httpResponseBody": True,
+                "geolocation": "US",
+                "echoData": 123,
+            }
+        },
+        {"zyte_api": {"httpResponseBody": True, "randomParameter": None}},
```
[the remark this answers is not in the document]
Gallaecio: This test does not cover automap usage, so I don’t think that is necessary.

Moreover, I think the point of the test is to verify how API responses are mapped into Scrapy-like responses, and I am thinking that we could remove most of the scenarios here, and only leave 1 for browserHtml and 1 for httpResponseBody. How parameter combinations reach the API is tested elsewhere already.

tests/test_api_requests.py
Gallaecio: I believe it is technically possible for it to be. If it was not technically possible, it should be Scrapy itself complaining, not this plugin.

I am not sure whether or not Zyte API itself allows it, but if it does not, I think it may be best to let Zyte API be the one who complains; similarly to how I think we should allow `httpResponseBody` and `browserHtml` be combined, even though Zyte API does not currently support that.

tests/test_api_requests.py
Gallaecio: I am assuming that Zyte API may at some point in the future allow combining browserHtml and httpResponseBody, and if we make that assumption, in this specific scenario, there should be no warning, because `httpResponseBody` is also set to `True`. Notice that in this scenario it is not the user passing `customHttpRequestHeaders`, that value comes from automated mapping.

I have no strong opinion on supporting combining browserHtml and httpResponseBody, though. But then we would have to decide which one to send when both are defined. And we may need to revisit what to do when none of the known current main outputs (httpResponseBody, browserHtml, screenshot) are `True`.

scrapy_zyte_api/handler.py
kmike: > Naming-wise, don’t you prefer ZYTE_API_AUTOMAP for this, or something in those lines (e.g. TRANSPARENT_MODE, PROXY_MODE)? To me DOWNLOAD feels generic enough that it could be confusing, since you can “download” content with DOWNLOAD=False, i.e. someone that still does not know what it does would wonder how ENABLED and DOWNLOAD are different.

Naming.. :) My thinking was: we should have a name which communicates that "turn this option ON - now all your downloads are handled by Zyte API". `..PROXY_MODE` was my first thought, but it might get a bit confusing, as Zyte API is not a proxy now (and it's possible it'd get a proxy interface later).  I'm fine with `ZYTE_API_TRANSPARENT_MODE` or some other name. "AUTOMAP" sounds more low-level. 

By the way, in the proposal above, it's a single option, which enables automatic mapping and enables Zyte API on all requests. But we shuld have a way to disable it per-request, I haven't thought about it.

> I am not sure what you mean here, it sounds OK to me.

I'm not fond of `browserHtml` parameter behaving very different from other parameters in `zyte_api_extra`. For others, you can just add them to API request, but `browserHtml` changes the behavior completely. But it might be ok.

Gallaecio: > 1. Add an option like `ZYTE_API_DOWNLOAD=True`. This enables Mode (1) - automatic mapping of scrapy.Request, etc. For the user it looks like everything works as before, but Zyte API is now used to download data.

The way you describe it, it sounds to me like `ZYTE_API_TRANSPARENT_MODE` would not only activate automated mapping of request fields, but also implicitly set `ZYTE_API_ON_ALL_REQUESTS=True`.

At first I thought you wanted to merge both settings, but I think a setting like `ZYTE_API_ON_ALL_REQUESTS` still has its place, specially in combination with `ZYTE_API_DEFAULT_PARAMS`, so I am not sure what the best way forward is here. `settings.get("ZYTE_API_ON_ALL_REQUESTS", settings.get("ZYTE_API_TRANSPARENT_MODE", False))`?

I also wonder, if users want automated mapping for specific requests, what would the `meta` look like? `"zyte_api_transparent_mode": True, "zyte_api_extra": {"browserHtml": True}`? I think it would be nice if you did not need 2 keys for this, but having `"zyte_api_extra"` with a value other than `False` enable transparent mode feels wrong naming-wise. I wonder if `"zyte_api_transparent_mode": {"browserHtml": True}` would be OK.

kmike: > The way you describe it, it sounds to me like ZYTE_API_TRANSPARENT_MODE would not only activate automated mapping of request fields, but also implicitly set ZYTE_API_ON_ALL_REQUESTS=True

Yes, kind-of. I was thinking there is no really a need for `ZYTE_API_ON_ALL_REQUESTS=True` if user uses "explicit" mode (data extraction, screenshots) - in these cases you'd likely want to enable things manually. If you don't do it, you can get screenshots of sitemaps, or attempted data extraction from robots.txt or media files. 

If `ZYTE_API_TRANSPARENT_MODE=True`, then I don't see a use case for disabling `ZYTE_API_ON_ALL_REQUESTS`. So, `ZYTE_API_ON_ALL_REQUESTS` doesn't look vary useful if we have transparent mode.

With that, it seems the main question is how ZYTE_API_DEFAULT_PARAMS behaves. We have 2 options:

1. ZYTE_API_DEFAULT_PARAMS are applied both for "transparent" and "explicit" requests. For example, one may want to set geolocation for all requests.
2. ZYTE_API_DEFAULT_PARAMS is only applied for "explicit" `meta={"zyte_api": ...}` requests. For example, someone may want to set screenshot options.

Thoughts?

> I also wonder, if users want automated mapping for specific requests, what would the meta look like?

Hm, `"zyte_api_transparent_mode": {"browserHtml": True}` makes sense, but not we have 2 way to do the same - 2nd is via `zyte_api_extra`. Or do yuou propose not to use `"zyte_api_extra"` name? This could work. But ``"zyte_api_transparent_mode"` name is very long; unlike function names, this is actually an issue, because it's easy to make a typo, and it's much harder to get auto-completion.

tests/test_api_requests.py
Gallaecio: > In any case, I see that there is no test covering the case where a user sets `browserHtml=True` and `customHttpRequestHeaders={…}`, but not `httpResponseBody`. I will extend tests accordingly.

[This is done now](https://github.com/scrapy-plugins/scrapy-zyte-api/pull/41/commits/1bad96bb9b5a3664f24a778e68e435004abb759e).

tests/test_api_requests.py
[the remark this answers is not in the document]
Gallaecio: > I think the point of the test is to verify how API responses are mapped into Scrapy-like responses, and I am thinking that we could remove most of the scenarios here, and only leave 1 for browserHtml and 1 for httpResponseBody. How parameter combinations reach the API is tested elsewhere already.

[This is done now](https://github.com/scrapy-plugins/scrapy-zyte-api/pull/41/commits/c59fb8bf828816ffdc6833dfa1373764ba02493b).

scrapy_zyte_api/handler.py
Gallaecio: OK, so what about this:
- `ZYTE_API_ENABLED` (default: `True`) toggles the plugin, but has no effect beyond that: requests still do not go through Zyte Data API unless request metadata demands so. This is mostly a way to allow easily disabling the plugin.
- Request metadata key `zyte_api` enables Zyte Data API for a request with the request URL and the specified parameters, merged with `ZYTE_API_DEFAULT_PARAMS`.
- There is no way to make all requests go through Zyte Data API with `ZYTE_API_DEFAULT_PARAMS`. We can consider adding it in the future if we see a valid use case. The workaround would be to set `"zyte_api": True` on all requests.
- Request metadata key `zyte_api_automap` enables Zyte Data API for a request with automated mapping, and can be used to pass additional parameters for that mode. Combining it with `zyte_api` causes an error. It is merged with `ZYTE_API_AUTOMAP_PARAMS`. 
- `ZYTE_API_TRANSPARENT_MODE` (default: `False`) makes it so that by default all requests are sent through Zyte Data API with automatic request parameter mapping enabled by default. Request metadata key `zyte_api_automap` can be used to provide extra parameters, or to disable automatic mapping for specific requests (and Zyte Data API usage, unless `zyte_api` is also used). Use of `zyte_api` on a request implies `"zyte_api_automap": False` for that request.

I wonder if we should deprecate and replace `zyte_api` and `ZYTE_API_DEFAULT_PARAMS` with `zyte_api_raw` and `ZYTE_API_RAW_PARAMS`. I think “raw” may help in fighting people’s expectations of automated parameter mapping.

We could also consider replacing `ZYTE_API_TRANSPARENT_MODE` with an enum-based `ZYTE_API_MODE` supporting values `manual` (default) and `transparent`, which we could later extend to support an additional value (e.g. `all`, or `automatic`) causing all requests to go through Zyte Data API with automated mapping disabled by default, i.e. using `ZYTE_API_DEFAULT_PARAMS`.

kmike: >   ZYTE_API_ENABLED (default: True) toggles the plugin, but has no effect beyond that: requests still do not go through Zyte Data API unless request metadata demands so. This is mostly a way to allow easily disabling the plugin.

Makes sense.


> Request metadata key zyte_api enables Zyte Data API for a request with the request URL and the specified parameters, merged with ZYTE_API_DEFAULT_PARAMS.

Makes sense. 

> There is no way to make all requests go through Zyte Data API with ZYTE_API_DEFAULT_PARAMS. We can consider adding it in the future if we see a valid use case. The workaround would be to set "zyte_api": True on all requests.

+1; we can consider it later if there is a use case. It's also trivial for the users to write a middleware which does it.

> Request metadata key zyte_api_automap enables Zyte Data API for a request with automated mapping, and can be used to pass additional parameters for that mode. Combining it with zyte_api causes an error. It is merged with ZYTE_API_AUTOMAP_PARAMS.

Depending on the implementation, it may contradict the next point. But overall, it sounds good.

> ZYTE_API_TRANSPARENT_MODE (default: False) makes it so that by default all requests are sent through Zyte Data API with automatic request parameter mapping enabled by default. Request metadata key zyte_api_automap can be used to provide extra parameters, or to disable automatic mapping for specific requests (and Zyte Data API usage). Use of zyte_api on a request implies "zyte_api_automap": False for that request.

Makes sense.

> I wonder if we should deprecate and replace zyte_api and ZYTE_API_DEFAULT_PARAMS with zyte_api_raw and ZYTE_API_RAW_PARAMS. I think “raw” may help in fighting people’s expectations of automated parameter mapping.

That's reasonable, but more verbose. I think if we go this way, we need to think how can we simplify non-transparent requests further (browserHtml, data extraction).

Initially I had a design similar to scrapy-splash in mind:

* zyte_api meta key is the source of truth. It provides a raw API request (see https://github.com/scrapy-plugins/scrapy-zyte-api/issues/48), maybe without "url" though, not sure.
* Automatic mapping works by updating zyte_api meta key.
* Transparent mode works by updating zyte_api meta key.
* To support automatic mapping wihout transparent mode, and to simplify request building (no typos in string meta keys), we introduce a custom ZyteApiRequest.

I'm not sure it's the best though, given the issues we had with scrapy-splash implementation - it's quite hard to follow (partially because of mutability), and it lead to security issues. It also doesn't solve the problem of disabling individual requests when transparent mode is enabled, though it's solvable. Two separate meta keys may be fine. 

> We could also consider replacing ZYTE_API_TRANSPARENT_MODE with an enum-based ZYTE_API_MODE supporting values manual (default) and transparent, which we could later extend to support an additional value (e.g. all, or automatic) causing all requests to go through Zyte Data API with automated mapping disabled by default, i.e. using ZYTE_API_DEFAULT_PARAMS.

This sounds ok, thought I don't see big benefits.

Gallaecio:
@kmike @BurnzZ Ready for a new review round!

tests/test_api_requests.py
```diff
@@ -322,6 +233,1467 @@ async def parse(self, response):
...
+            {"browserHtml": True, "httpResponseHeaders": True, "screenshot": True},
+            [],
+        ),
+        # If no known main output is specified, and httpResponseBody is
+        # explicitly set to False, httpResponseBody is unset and no main output
+        # is added.
+        (
+            {"httpResponseBody": False},
+            {},
+            [],
```
BurnzZ, line 690: Should we log a warning that the user is not sending any meaningful payload to Zyte API?

tests/test_api_requests.py
```diff
@@ -322,6 +233,1467 @@ async def parse(self, response):
...
+                "screenshot": True,
+            },
+            [],
+        ),
+        # If httpResponseBody is True, implicitly or explicitly, and there is
+        # no other known main output parameter (browserHtml, screenshot),
+        # Request.headers are mapped as customHttpRequestHeaders only.
+        #
+        # While future main output parameters are likely to use requestHeaders
+        # instead, we cannot known if an unknown parameter is a main output
```
BurnzZ, line 1040:
```suggestion
        # instead, we cannot know if an unknown parameter is a main output
```

tests/test_api_requests.py
Gallaecio: I don’t think so.

Zyte API will currently fail with a bad request error, which achieves the same goal, I think. And if in the future this becomes a valid request, then the warning would be wrong.

On the other hand, if the way they make this a valid request in the future is to make `"httpResponseBody": True` the default in this case, then the automapped parameters (none) would seemingly contradict user intent (`"httpResponseBody": False`). But I think the only scenario where a user sets `"httpResponseBody": False` is one where the user includes some other (currently unknown) parameter that indicates a different output, and I don’t expect Zyte API to ever make `"httpResponseBody": True` implicit in such cases.

README.rst
```diff
@@ -60,12 +66,49 @@ Here's an example of the things needed inside a Scrapy project's ``settings.py``
...
+
+You can send a request through Zyte API in one of the following ways:
+
+-   Setting all Zyte API parameters manually, keeping full control of what is
+    sent to Zyte API. See **Sending requests with manually-defined parameters**
+    below.
+
+-   Letting Zyte API parameters be chosen automatically based on your Scrapy
+    request parameters where possible. See **Sending requests with
+    automatically-mapped parameters** below.
```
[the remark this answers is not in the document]
Gallaecio: No opinion either way from me.

I think there are valid reasons to prefer raw parameters, but we also know that people can make easy mistakes when using raw parameters (e.g. not enabling `httpResponseHeaders`).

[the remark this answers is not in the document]
kmike: I think it'd be good to document it in the following order (in the intro, but later in the docs as well) - at least from "getting started" point of view:

1) Transparent mode.
2) Manual mode.
3) Automatic mapping without transparent mode.

README.rst
```diff
@@ -87,155 +129,156 @@ must be present and contain a dict with Zyte API parameters:
...
 
-You can see the full list of parameters in the `Zyte API Specification
-<https://docs.zyte.com/zyte-api/openapi.html#zyte-openapi-spec>`_.
-The ``url`` parameter is filled automatically from ``request.url``, other 
-parameters should be set explicitly.
+See the `Zyte API documentation`_ to learn about Zyte API parameters.
 
-The raw Zyte API response can be accessed via the ``raw_api_response``
-attribute of the response object.
+.. _Zyte API documentation: https://docs.zyte.com/zyte-api/get-started.html
```
kmike, line 141: It seems that from the docs there are 2 sections which are relevant for scrapy-zyte-api users who want to know about Zyte API parameters:

* https://docs.zyte.com/zyte-api/usage/extract.html and
* https://docs.zyte.com/zyte-api/openapi.html

README.rst
```diff
@@ -87,155 +129,156 @@ must be present and contain a dict with Zyte API parameters:
...
 
-When you use the Zyte API parameters ``browserHtml``, ``httpResponseBody``, or
-``httpResponseHeaders``, the response body and headers are set accordingly.
 
-Note that, for Zyte API requests, the spider gets responses of
-``ZyteAPIResponse`` and ``ZyteAPITextResponse`` types,
-which are respectively subclasses of ``scrapy.http.Response``
-and ``scrapy.http.TextResponse``.
+Sending requests with automatically-mapped parameters
+-----------------------------------------------------
```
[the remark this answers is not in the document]
kmike: As mentioned earlier, I'd probably start with `ZYTE_API_TRANSPARENT_MODE`, even before the manual mode.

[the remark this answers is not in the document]
kmike: Though that's not a strong opinon.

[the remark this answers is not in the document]
kmike: The reason is that most people should be using transparent mode, not `"zyte_api_automap": True`.

README.rst
```diff
@@ -87,155 +129,156 @@ must be present and contain a dict with Zyte API parameters:
...
+Zyte API header support changes in the future, and you cannot upgrade to a
+version of scrapy-zyte-api that includes such changes, use the following
+settings to change which headers are mapped and how they are mapped:
 
-        custom_settings = {
-            "ZYTE_API_DEFAULT_PARAMS": {
-                "geolocation": "US",  # You can set any Geolocation region you want.
-            }
-        }
+-   ``ZYTE_API_UNSUPPORTED_HEADERS`` determines headers that *cannot* be mapped
```
kmike, line 184: I wonder if there are other reasons to skip the headers, not only that API doesn't support them. E.g. user-agent, or cookies, not sure. If there are such reasons, it could make sense to rename the option to something like ZYTE_API_SKIP_HEADERS, as the headers may be supported, but user still may want to avoid sending them.

README.rst
```diff
@@ -87,155 +129,156 @@ must be present and contain a dict with Zyte API parameters:
...
 
-    import scrapy
+-   ``Request.method`` becomes ``httpRequestMethod`` even for unsupported_
+    ``httpRequestMethod`` values, and even if ``httpResponseBody`` is unset.
 
+    .. _unsupported: https://docs.zyte.com/zyte-api/usage/extract.html#zyte-api-set-method
 
-    class SampleQuotesSpider(scrapy.Spider):
-        name = "sample_quotes"
+-   You can set ``customHttpRequestHeaders`` or ``requestHeaders`` to ``True``
```
kmike, line 209: What do you think about adding a few examples? The reason I'm asking is that by this time it might not be clear where to set requestHeaders to True. Is it `settings`, `request.meta`, `request.meta['zyte_api']` or `request.meta['zyte_api_automap']`?

README.rst
```diff
@@ -87,155 +129,156 @@ must be present and contain a dict with Zyte API parameters:
...
+-   ``ZYTE_API_AUTOMAP_PARAMS`` is a ``dict`` of parameters to be combined with
+    automatically-mapped parameters. See **Sending requests with
+    automatically-mapped parameters** above.
+
+For example, if you set ``ZYTE_API_DEFAULT_PARAMS`` to
+``{"geolocation": "US"}`` and ``zyte_api`` to ``{"browserHtml": True}``,
+``{"url: "…", "geolocation": "US", "browserHtml": True}`` is sent to Zyte API.
+
+Parameters in these settings are merged with request-specific parameters, with
+request-specific parameters taking precedence.
```
kmike, line 277: It'd be nice to clarify behavior of these options for the requests which are not marked as manual or auto-mapped zyte API requests.

scrapy_zyte_api/handler.py
```diff
@@ -21,11 +26,485 @@
...
+    "browserHtml": False,
+    "screenshot": False,
+}
+
+
+def _iter_headers(
+    *,
+    api_params: Dict[str, Any],
+    request: Request,
+    parameter: str,
```
kmike, line 39: header_parameter?

scrapy_zyte_api/handler.py
```diff
@@ -21,11 +26,485 @@
...
+        return
+    for k, v in request.headers.items():
+        if not v:
+            continue
+        decoded_v = b",".join(v).decode()
+        lowercase_k = k.strip().lower()
+        yield k, lowercase_k, decoded_v
+
+
+def _map_custom_http_request_headers(
```
kmike, line 58: What do you think about moving automap-related utilities to a separate module? 

I was thinking about a function like this (I think you have something like this already):

```py
def get_zyte_api_request(request: scrapy.Request, zyte_automap: dict) -> dict[str, Any]:
    """ Get a scrapy.Request, as well as automap params (already merged with default params). 
    Return parameters for Zyte API request. """
```

tests/__init__.py
```diff
@@ -29,15 +30,19 @@ async def make_handler(settings: dict, api_url: Optional[str] = None):
...
-        crawler=crawler,
-    )
+    try:
+        handler = create_instance(
+            ScrapyZyteAPIDownloadHandler,
+            settings=None,
+            crawler=crawler,
+        )
+    except NotConfigured:
+        handler = None
```
kmike, line 40: could you please add a comment, to explain when this is happening?

tests/test_api_requests.py
```diff
@@ -322,6 +233,1467 @@ async def parse(self, response):
...
+                },
+                ["Use Request.method"],
+            )
+            for request_method, meta_method in (
+                ("GET", "GET"),
+                ("POST", "POST"),
+            )
+        ),
+        # If httpRequestMethod is also specified in meta with a different value
+        # from Request.method, a warning is logged asking to use Request.meta,
```
kmike, line 903: Request.method, not Request.meta?

tests/test_responses.py
```diff
@@ -43,6 +44,20 @@ def raw_api_response_body():
...
 
+def raw_api_response_mixed():
+    return {
+        "url": URL,
+        "browserHtml": PAGE_CONTENT,
+        "httpResponseBody": b64encode(PAGE_CONTENT_2.encode("utf-8")),
+        "echoData": {"some_value": "here"},
+        "httpResponseHeaders": [
+            {"name": "Content-Type", "value": "text/html"},
+            {"name": "Content-Length", "value": len(PAGE_CONTENT)},
```
kmike, line 55: it's more likely to match the length of PAGE_CONTENT_2

kmike (approved):
Great job @Gallaecio! 

I've left some minor comments and suggestions, but I'm fine with merging it as-is. The PR looks very good.

tests/test_api_requests.py
Gallaecio: I think the wording is correct, however confusing. It is saying “if Request.meta["httpRequestMethod"] != Request.method”.

kmike: Hm, but the warning asks to use Request.method, not Request.meta.

scrapy_zyte_api/handler.py
```diff
@@ -21,11 +26,485 @@
...
+                "neither browserHtml nor screenshot are set to True."
+            )
+        api_params.setdefault("httpResponseHeaders", True)
+    elif (
+        api_params.get("httpResponseHeaders") is False
+        and not default_params.get("httpResponseHeaders") is False
+    ):
+        logger.warning(
+            "You do not need to set httpResponseHeaders to False if "
+            "you do set httpResponseBody or browserHtml to True. Note "
```
BurnzZ, line 212:
```suggestion
            "httpResponseBody or browserHtml isn't set to True. Note "
```

tests/test_api_requests.py
```diff
@@ -322,6 +233,1467 @@ async def parse(self, response):
...
+            },
+            ["do not need to set httpResponseHeaders to False"],
+        ),
+    ],
+)
+def test_automap_default_parameter_cleanup(meta, expected, warnings, caplog):
+    _test_automap({}, {}, meta, expected, warnings, caplog)
+
+
+# @pytest.mark.xfail(reason="To be implemented", strict=True)
```
BurnzZ, line 1661: Should we remove this comment?

tests/test_api_requests.py
```diff
@@ -322,6 +233,1467 @@ async def parse(self, response):
...
+        # If no known main output is specified, and httpResponseBody is
+        # explicitly set to False, httpResponseBody is unset and no main output
+        # is added.
+        (
+            {"httpResponseBody": False},
+            {},
+            [],
+        ),
+        (
+            {"httpResponseBody": False, "unknownMainOutput": True},
```
BurnzZ, line 693: I may have missed it as you may have a succinct test somewhere but to make sure, are there tests using `None` unset Zyte API params like these?

Gallaecio: There is `test_default_params_merging` to test that a param from a default params setting can be unset with `None`. But if you mean to test that `param: None` and `param: False` behave the same, there is no test.

Should we allow `None` to have the same effect as `False`? (no opinion either way from me, and I am not sure that is what you meant here)

BurnzZ: I was thinking of a test that ensures `param: None` would behave the same way as if `param` wasn't set at all. Having such a test could help avoid potential regressions down the line.

README.rst
Gallaecio: I will be adding 2 code examples prior to this, to cover the default mapping behavior and the changes that e.g. `"screenshot": True` introduces.

But I am hesitant about including examples of this specific bullet point, or any sibling bullet point. These bullet points describe how automated parameter mapping responds to stuff that Zyte API does not currently support, and while I think it needs to be covered for completeness, I think it is OK if users do not fully understand this part, as they are not meant to be using it anyway.

When future Zyte API changes cause some of these bullet points to be relevant for actual Zyte API usage, then we can move those bullet points out of here into a different list, above, and include an example for them.

scrapy_zyte_api/handler.py
Gallaecio: I am moving `_get_api_params` into its own module, but I am not sure if this is what you had in mind.

Gallaecio:
@kmike I backtracked on what we discussed elsewhere about marking `httpResponseHeaders` as an implementation detail in the documentation, because depending on how Zyte API implements text decoding, it may be best that automatic parameter mapping continues to use the current `httpResponseBody` + `httpResponseHeaders` combo by default to support requests for binary data as well.

README.rst
```diff
@@ -147,52 +151,71 @@ taking priority.
...
-                url="http://quotes.toscrape.com/",
-                callback=self.parse,
+                url="https://quotes.toscrape.com/",
                 meta={
                     "zyte_api": {
-                        "browserHtml": True,
-                        "javascript": True,
-                        "echoData": {"some_value_I_could_track": 123},
+                        "httpResponseBody": True,
+                        "httpResponseHeaders": True,
```
kmike, line 192: I think that using browserHtml might be better here, as it shows a real use case for using the manual mode.

README.rst
```diff
@@ -60,85 +66,83 @@ Here's an example of the things needed inside a Scrapy project's ``settings.py``
...
-`Request.meta <https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.meta>`_
-must be present and contain a dict with Zyte API parameters:
+The ``ZYTE_API_ENABLED`` setting, which is ``True`` by default, can be set to
+``False`` to disable this plugin.
 
-.. code-block:: python
 
-    import scrapy
+Usage
+=====
```
kmike, line 74: I think there are 2 gotchas which we currently don't call out explicitly. 

1. If you use manual mode with httpResponseBody, you need to pass httpResponseHeaders=true for decoding to work properly. In general, Request parameter mapping is documented very well in the README, but I'm not sure if existing docs are enough for the Response parameter mapping. 
2. In the section about default parameters we should probably mention that using browserHtml=True (or anything other than httpResponseBody + httpResponseHeaders as the main output types) for all requests is probably a very bad idea, because it'd be applied for robots.txt, sitemap and other similar requests; it's not going to work correctly.

What do you think about adding them somewhere to the document?

README.rst
Gallaecio: I ended up covering both, ``browserHtml`` as a “normal” example, and this as an example of the need to include headers for proper decoding. Not sure if having the latter as an example is overkill.

README.rst
Gallaecio: Response mapping now has its own section with 2 examples and detailed parameter mapping. Also covered 2.

README.rst
```diff
@@ -402,8 +490,19 @@ Parameters in these settings are merged with request-specific parameters, with
...
+request parameter mapping, and ``ZYTE_API_AUTOMAP_PARAMS`` has no effect on
+requests that use manually-defined parameters.
+
+When using transparent mode (see **Using transparent mode** above), be careful
+of which parameters you define through ``ZYTE_API_AUTOMAP_PARAMS``. In
+transparent mode, all Scrapy requests go through Zyte API, even requests that
+Scrapy sends automatically, such as those for ``robots.txt`` files when
+ROBOTSTXT_OBEY_ is ``True``, or those for sitemaps when using a `sitemap
+spider`_. Certain parameters, like ``browserHtml`` or ``screenshot``, are not
+meant to be used for every single request.
```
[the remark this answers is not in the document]
Gallaecio: Which example? Do you mean this one?

> For example, if you set ``ZYTE_API_DEFAULT_PARAMS`` to ``{"geolocation": "US"}`` and ``zyte_api`` to ``{"browserHtml": True}``, ``{"url: "…", "geolocation": "US", "browserHtml": True}`` is sent to Zyte API.

Gallaecio:
I have documented `httpResponseBody` and `httpResponseHeaders` being `True` by default when using automatically-mapped request parameters as an implementation, recommended enabling them manually in the right scenarios, and removed warnings that used to be triggered by that.

I also realized that there was no reason for `httpResponseHeaders` to be enabled as a side effect of enabling `browserHtml`, and changed that as well. I suspect it stemmed from me misunderstanding how the current implementation works (i.e. maybe I was thinking without response headers `browserHtml` response were also treated as binary responses).

README.rst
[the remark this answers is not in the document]
kmike: yes, this one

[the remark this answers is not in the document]
kmike: Makes sense!

kmike (approved):
Looks awesome, thanks @Gallaecio!

kmike:
Ok, there are no further comments, so let's merge it :)

PR #49, opened 2022-09-12
Extend Scrapy support to version 2.0.1

Gallaecio:


.pre-commit-config.yaml
```diff
@@ -18,4 +18,6 @@ repos:
     rev: v0.971
     hooks:
     - id: mypy
+      additional_dependencies:
+      - types-setuptools
```
Gallaecio, line 22: Needed due to the use of `pkg_resources`.

.github/workflows/test.yml
```diff
@@ -30,7 +37,7 @@ jobs:
         python -m pip install tox
     - name: tox
       run: |
-        tox -e py
+        tox -e ${{ matrix.toxenv  || 'py' }}
```
BurnzZ, line 40:
```suggestion
        tox -e ${{ matrix.toxenv || 'py' }}
```

BurnzZ (approved):
LGTM @Gallaecio !

BurnzZ:
Ohhh I forgot to add the we also need to update the `README.rst` file to reflect the new minimum Scrapy version.

scrapy_zyte_api/responses.py
```diff
@@ -20,11 +22,27 @@ class ZyteAPIMixin:
...
+            self.attributes: Tuple[str, ...] = (
+                "url",
+                "status",
+                "headers",
+                "body",
+                "request",
+                "flags",
+                "certificate",
+            )
+            if parse_version(scrapy.__version__) >= parse_version("2.1.0"):
```
kmike, line 35: That's quite minor, but could you please move Scrapy version check code to a separate function (e.g. `utils.scrapy_210_or_later`, or `scrapy_supports_attributes_ip_address`)? It would help with code duplication a tiny bit (there are similar checks in tests), and would allow to isolate pkg_resources import, which use is discouraged now (https://setuptools.pypa.io/en/latest/pkg_resources.html). 

As for pkg_resources, it seems the replacement for parse_version is not available in the suggested replacement modules, so that look fine to keep using it.

scrapy_zyte_api/responses.py
```diff
@@ -75,11 +96,12 @@ def from_api_response(cls, api_response: Dict, *, request: Request = None):
             raw_api_response=api_response,
         )
 
+    def replace(self, *args, **kwargs):
+        kwargs.setdefault("encoding", self.encoding)
+        return ZyteAPIMixin.replace(self, *args, **kwargs)
```
kmike, line 101: It seems super() is going to work here, but I don't see an issue with having explicit ZyteAPIMixin as well.

kmike:
Thanks @Gallaecio for the careful investigation!

PR #50, opened 2022-09-16
Zyte Data API → Zyte API

Gallaecio:
Note: Ignored `CHANGES.rst` on purpose, but no strong opinion on that, I can change as well if desired.


BurnzZ (approved):
LGTM! Though I think we could update `CHANGES.rst` as well to make the naming more consistent.

kmike:
Thanks @Gallaecio!

PR #51, opened 2022-09-20
Prepare for 0.5.1 release

BurnzZ:
Requires https://github.com/zytedata/python-zyte-api/pull/32 to be released to PyPI first.

I'm proposing for a `0.5.1` release instead of a `0.6.0` since it's a simply dependency update. Even though we're now enforcing brotli codecs under the hood, in Scrapy's POV, it's a small update and doesn't change anything on how developers directly interact with Zyte API.

TODO after merge:
- Tag `0.5.1` to latest master branch and push.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Changes
 =======
 
+0.5.1 (2022-09-20)
+------------------
+
+* Raise the minimum dependency of Zyte API's Python API to ``zyte-api>=0.4.0``.
+  This changes all the requests to Zyte API to have have ``Accept-Encoding: br``
+  and automatically decompress brotli responses.
+
 0.5.0 (2022-08-25)
 ------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Changes
 * Raise the minimum dependency of Zyte API's Python API to ``zyte-api>=0.4.0``.
   This changes all the requests to Zyte API to have have ``Accept-Encoding: br``
   and automatically decompress brotli responses.
+* Rename "Zyte Data API" to simply "Zyte API" in the README.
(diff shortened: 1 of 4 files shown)


Gallaecio:
I don’t think we need to change the minimum version or release a new version of scrapy-zyte-api. The current version should work with both 0.3 and 0.4 versions of python-zyte-api.

CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Changes
 =======
 
+0.5.1 (2022-09-20)
+------------------
+
+* Raise the minimum dependency of Zyte API's Python API to ``zyte-api>=0.4.0``.
+  This changes all the requests to Zyte API to have have ``Accept-Encoding: br``
+  and automatically decompress brotli responses.
```
kmike, line 9: There is also https://github.com/scrapy-plugins/scrapy-zyte-api/pull/49 in the main branch

kmike:
> I don’t think we need to change the minimum version or release a new version of scrapy-zyte-api. The current version should work with both 0.3 and 0.4 versions of python-zyte-api.

Do we want it though? Force-upgrading in this case could be a good thing to do, as it allows to reduce client-server traffic. There could be an issue if the new client library version doesn't work for some reason, but in this case we should fix it.

Gallaecio:
If we go ahead, we need to update https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/tox.ini#L22 for tests with frozen dependencies to pass.

CHANGES.rst
BurnzZ: Good catch! Added them in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/51/commits/d983a7563b55f66eebe60d3d749ac84ce8aea2e2.

Gallaecio (approved):

kmike:
Thanks @BurnzZ!

PR #56, opened 2022-10-26
Provide a request fingerprinter for Scrapy 2.7+

Gallaecio:


README.rst
```diff
@@ -615,3 +629,41 @@ Stats
...
+
+If you set the ``REQUEST_FINGERPRINTER_CLASS`` setting to
+``"scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter"``, `request
+fingerprinting <https://docs.scrapy.org/en/latest/topics/request-response.html#request-fingerprints>`_
+behaves as follows:
+
+-   Requests sent through Zyte API get a fingerprint based on the following
+    Zyte API parameters:
+
+    -   ``url`` (`canonicalized <https://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.canonicalize_url>`)
```
BurnzZ, line 645:
```suggestion
    -   ``url`` (:ref:`canonicalized <https://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.canonicalize_url>`)
```

README.rst
```diff
@@ -48,27 +48,41 @@ in the ``settings.py`` of your Scrapy project.
...
     TWISTED_REACTOR = "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
 
+    # Only on Scrapy 2.7 or later:
+    REQUEST_FINGERPRINTER_CLASS = "scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter"
+
 The ``ZYTE_API_ENABLED`` setting, which is ``True`` by default, can be set to
 ``False`` to disable this plugin.
 
+If you had a value for the ``REQUEST_FINGERPRINTER_CLASS`` previously set,
+assign that value to the ``ZYTE_API_FALLBACK_REQUEST_FINGERPRINTER_CLASS``
```
BurnzZ, line 82: Let's show an example on how to assign this to `ZYTE_API_FALLBACK_REQUEST_FINGERPRINTER_CLASS` so that users can easily copy it from the code block above.

README.rst
```diff
@@ -615,3 +629,41 @@ Stats
...
+-   Requests sent through Zyte API get a fingerprint based on the following
+    Zyte API parameters:
+
+    -   ``url`` (`canonicalized <https://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.canonicalize_url>`)
+
+    -   Request attribute parameters (``httpRequestBody``,
+        ``httpRequestMethod``)
+
+    -   Output parameters (``browserHtml``, ``httpResponseBody``,
+        ``httpResponseHeaders``, ``screenshot``)
```
BurnzZ, line 651:
```suggestion
        ``httpResponseHeaders``, ``screenshot``, etc.)
```

Lots more to come next year with Automated Extraction.

scrapy_zyte_api/_request_fingerprinter.py
```diff
@@ -0,0 +1,81 @@
+from typing import TYPE_CHECKING
+
+try:
+    from scrapy.utils.request import RequestFingerprinter  # NOQA
+except ImportError:
+    if not TYPE_CHECKING:
+        ScrapyZyteAPIRequestFingerprinter = None
+else:
```
BurnzZ, line 8: I'm thinking of ways if we could simplify the configuration and setup on this regard. Currently, users need to set the following if they have Scrapy 2.7+
- `REQUEST_FINGERPRINTER_CLASS`
- `ZYTE_API_FALLBACK_REQUEST_FINGERPRINTER_CLASS`

If users are using Scrapy versions below this, they won't need to set it. But if they do upgrade in the future, it's another step that they need to consider. Probably even forget about it.

What do you think about setting `REQUEST_FINGERPRINTER_CLASS="scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter"` by default regardless of what Scrapy version they use?

But in this case, `ScrapyZyteAPIRequestFingerprinter` automatically detects which Scrapy version the user is using and changes its behavior accordingly. So when users do upgrade Scrapy versions in the future, it's one less setting modification that they need to worry about.

One downside here though is that such convenience could remove the user's awareness that the Fingerprinting methodology has changed.

README.rst
Gallaecio: Probably best to append them later. We can then figure out how to refer to them, i.e. we may not want to list them all here.

scrapy_zyte_api/_request_fingerprinter.py
Gallaecio: Setting the setting on older Scrapy versions is actually already possible. We could update the documentation not to mention the 2.7 requirement, or to clarify that the setting should be set, but it will only be effective on Scrapy 2.7 and later.

BurnzZ: I see, that's awesome! 

+1 on this approach since the user can just set and forget it. No need to revisit and update the fingerprinter classes later on when they upgrade Scrapy.

README.rst
```diff
@@ -615,3 +619,109 @@ Stats
...
+If you use Scrapy 2.6.3 or earlier, Zyte API parameters are not taken into
+account for request fingerprinting. This can cause some Scrapy components, like
+the filter of duplicate requests or the HTTP cache extension, to interpret 2
+different requests as being the same.
+
+To avoid most issues, use automated request parameter mapping, either through
+transparent mode or setting ``zyte_api_automap`` to ``True`` in
+``Request.meta``, and then use ``Request`` properties attributes instead of
+``Request.meta`` as much as possible. Unlike ``Request.meta``, ``Request``
+attributes do affect request fingerprints in Scrapy 2.6.3 and earlier.
```
kmike, line 692: a nice catch

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,291 @@
...
+        ),
+        (
+            "https://example.com/?1",
+            "https://example.com/?2",
+            False,
+        ),
+        (
+            "https://example.com/#1",
+            "https://example.com/#2",
+            True,
```
kmike, line 275: This is actually an interesting point. I can see how it's compatible with the existing approach, and how it works well intransparent mode. But I wonder if we should take fragment in account if browser-related output is requested. Not sure if we can make it a default though, and if it's not a default, it can be introduced later.

Gallaecio: > But I wonder if we should take fragment in account if browser-related output is requested.

That is a great point. And I think it may be best to address it as part of this pull request.

README.rst
Gallaecio: Note to self: fix typo: “properties attributes” → attributes

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,371 @@
...
+        ScrapyZyteAPIRequestFingerprinter, settings=crawler.settings, crawler=crawler
+    )
+
+    request1 = Request("https://example.com", meta={"zyte_api": {"echoData": "foo"}})
+    request2 = Request("https://example.com", meta={"zyte_api": True})
+
+    fingerprint1 = job_fingerprinter.fingerprint(request1)
+    fingerprint2 = no_job_fingerprinter.fingerprint(request2)
+    fingerprint3 = job_fingerprinter.fingerprint(request1)
+    fingerprint4 = no_job_fingerprinter.fingerprint(request2)
```
BurnzZ, line 200: Could we combine `fingerprint1` and `fingerprint3` together? as well as `fingerprint2` and `fingerprint4`? as they are essentially the same.

Or was the intent to have a permutation like this?
```python
    fingerprint1 = job_fingerprinter.fingerprint(request1)
    fingerprint2 = no_job_fingerprinter.fingerprint(request2)
    fingerprint3 = job_fingerprinter.fingerprint(request2)
    fingerprint4 = no_job_fingerprinter.fingerprint(request1)
```

README.rst
```diff
@@ -615,3 +619,99 @@ Stats
...
+
+Zyte API parameters that affect request fingerprinting
+------------------------------------------------------
+
+The request fingerprinter class of this plugin generates request fingerprints
+for Zyte API requests based on the following Zyte API parameters:
+
+-   ``url`` (`canonicalized <https://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.canonicalize_url>`_)
+
+    URL canonicalization keeps the URL fragment, if any, if
```
BurnzZ, line 644:
```suggestion
    URL canonicalization keeps any fragments, in cases where
```

I needed to re-read this a couple of times since the first `if` creates some slight confusion.

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,371 @@
...
+        (
+            "https://example.com/?1",
+            "https://example.com/?2",
+            False,
+        ),
+        (
+            "https://example.com/#1",
+            "https://example.com/#2",
+            True,
+        ),
```
BurnzZ, line 276: A case like the following could also be great for completeness:
```python
        (
            "https://example.com/#1",
            "https://example.com/#",
            False,
        )
```

README.rst
```diff
@@ -615,3 +619,100 @@ Stats
...
+
+.. code-block:: python
+
+    ZYTE_API_FALLBACK_REQUEST_FINGERPRINTER_CLASS = "custom.RequestFingerprinter"
+
+By default, requests that do not go through Zyte API use the default request
+fingerprinter class of the installed Scrapy version.
+
+
+Request fingerprinting on Scrapy 2.6.3 and earlier
```
kmike, line 685: In the unlikely scenario we release 2.6.4 this would become imprecise, so probably it's better to use `< 2.7` instead.

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,386 @@
...
+            {},
+            False,
+        ),
+        (
+            {"httpResponseBody": False},
+            False,
+        ),
+        (
+            {"httpResponseHeaders": True},
+            False,
```
kmike, line 342: I think the expected behavior is True in this case

scrapy_zyte_api/_request_fingerprinter.py
```diff
@@ -0,0 +1,88 @@
...
+                skip_headers=self._skip_headers,
+                browser_headers=self._browser_headers,
+                job_id=None,
+            )
+            if api_params is not None:
+                has_no_body = api_params.get("httpResponseBody", False) is False
+                is_browser_request = any(
+                    api_params.get(key, False) for key in ("browserHtml", "screenshot")
+                )
+                keep_fragments = has_no_body or is_browser_request
```
kmike, line 79: what do you think about extracting a method like "_should_keep_fragment", with this logic?

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,386 @@
...
+        (
+            {"httpResponseBody": True, "screenshot": True},
+            False,
+        ),
+        (
+            {"browserHtml": True, "httpResponseBody": True, "screenshot": True},
+            False,
+        ),
+        (
+            {"unknownMainOutput": True},
```
kmike, line 369: It seems that if unknownMainOutput is browser-related output, this case has a proper behavior, but `{"unknownMainOutput": True, "httpResponseBody": True}` doesn't. And it's the opposite if unknownMainOutput is not a browser-related output: this one is not correct, and the other one is correct.

What do you think about being more conservative, and not taking fragment in account unless there is a known browser-related output set?

BurnzZ (approved):

kmike:
Thanks @Gallaecio!

PR #59, opened 2022-11-03
Use custom download slots that bypass AutoThrottle

### LINKED ISSUE #53: Bypass AutoThrottle
See if we can make it so that the AutoThrottle extension, when enabled, does not impact Zyte API requests, which handle website-specific throttling internally already.

Gallaecio:
Resolves #53.

Tested manually as well.

I see a lot of aspects that we might want to discuss:
- Should we rename the new downloader middleware after what it does (name suggestions?), or keep this generic name and try to concentrate any downloader middleware code that we may need in the future into this downloader middleware, instead of creating new downloader middlewares?
- Is it OK that the whole Zyte API parameter parsing is done just to determine whether or not Zyte API is used, or should we use a simpler, cheaper implementation for this new purpose?
- Is it OK that we are calling the Zyte API parameter parsing logic 3 times per request now? Should we cache those calls somehow? (I assume weak references to requests are not reliable here, as meta can change, and I am not sure what the best alternative would be)
- Is `zyte-api@` the right prefix for slot IDs? `zyte-api@example.com` looks cool to me, but I wonder if it can be misleading, or if `@` can be problematic.
- Is it OK that we are using private APIs of the Scrapy downloader? Should we try to make upstream to expose them as public APIs, be it before or after this change?
- Should I refactor `_params.py` to turn its functions into methods of its new class, to minimize parameter passing?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,8 +48,13 @@ To enable this plugin:
     <https://docs.scrapy.org/en/latest/topics/settings.html#std-setting-DOWNLOAD_HANDLERS>`_
     Scrapy setting to ``"scrapy_zyte_api.ScrapyZyteAPIDownloadHandler"``.
 
+-   Add ``"scrapy_zyte_api.ScrapyZyteAPIDownloaderMiddleware"`` to the
+    `DOWNLOADER_MIDDLEWARES
+    <https://docs.scrapy.org/en/latest/topics/settings.html#downloader-middlewares>`_
+    Scrapy setting with any value, e.g. ``1000``.
+
 -   Set the `REQUEST_FINGERPRINTER_CLASS
-    <https://docs.scrapy.org/en/latest/topics/request-response.html#request-fingerprinter-class>`
+    <https://docs.scrapy.org/en/latest/topics/request-response.html#request-fingerprinter-class>`_
     Scrapy setting to ``"scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter"``.
 
 -   Set the `TWISTED_REACTOR
@@ -69,6 +74,9 @@ For example, in the ``settings.py`` file of your Scrapy project:
     DOWNLOAD_HANDLERS = {
         "http": "scrapy_zyte_api.ScrapyZyteAPIDownloadHandler",
         "https": "scrapy_zyte_api.ScrapyZyteAPIDownloadHandler",
+    }
+    DOWNLOADER_MIDDLEWARES = {
+        "scrapy_zyte_api.ScrapyZyteAPIDownloaderMiddleware": 1000,
     }
     REQUEST_FINGERPRINTER_CLASS = "scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter"
     TWISTED_REACTOR = "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
--- a/scrapy_zyte_api/_downloader_middleware.py
+++ b/scrapy_zyte_api/_downloader_middleware.py
@@ -1,0 +1,32 @@
+from typing import Set
+
+from ._params import _ParamParser
+
+
+class ScrapyZyteAPIDownloaderMiddleware:
+
+    _slot_prefix = "zyte-api@"
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler)
+
+    def __init__(self, crawler) -> None:
+        self._param_parser = _ParamParser(crawler.settings)
+        self._handled_slots: Set[str] = set()
+        self._crawler = crawler
+
+    def process_request(self, request, spider):
+        if self._param_parser.parse(request) is None:
+            return
+
+        downloader = self._crawler.engine.downloader
+        slot_id = downloader._get_slot_key(request, spider)
+        if not slot_id.startswith(self._slot_prefix):
+            slot_id = f"{self._slot_prefix}{slot_id}"
+            request.meta["download_slot"] = slot_id
+        if slot_id in self._handled_slots:
+            return
+        _, slot = downloader._get_slot(request, spider)
+        slot.delay = 0
+        self._handled_slots.add(slot_id)
--- a/scrapy_zyte_api/_downloader_middleware.py
+++ b/scrapy_zyte_api/_downloader_middleware.py
@@ -1,5 +1,3 @@
-from typing import Set
-
 from ._params import _ParamParser
 
 
@@ -13,7 +11,6 @@ class ScrapyZyteAPIDownloaderMiddleware:
 
     def __init__(self, crawler) -> None:
         self._param_parser = _ParamParser(crawler.settings)
-        self._handled_slots: Set[str] = set()
         self._crawler = crawler
 
     def process_request(self, request, spider):
@@ -25,8 +22,5 @@ class ScrapyZyteAPIDownloaderMiddleware:
         if not slot_id.startswith(self._slot_prefix):
             slot_id = f"{self._slot_prefix}{slot_id}"
             request.meta["download_slot"] = slot_id
-        if slot_id in self._handled_slots:
-            return
         _, slot = downloader._get_slot(request, spider)
         slot.delay = 0
-        self._handled_slots.add(slot_id)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,9 +15,9 @@ repos:
     hooks:
         - id: flake8
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.971
+    rev: v0.991
     hooks:
     - id: mypy
       additional_dependencies:
       - types-setuptools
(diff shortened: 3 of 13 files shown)


wRAR:
> Should we rename the new downloader middleware after what it does (name suggestions?), or keep this generic name and try to concentrate any downloader middleware code that we may need in the future into this downloader middleware, instead of creating new downloader middlewares?

I think it only makes sense to split code into separate middewares if the user should be able to not enable some of them, to simplify subclassing them or if different code should be inserted into different positions on the mw list.

> Is zyte-api@ the right prefix for slot IDs? zyte-api@example.com looks cool to me, but I wonder if it can be misleading, or if @ can be problematic.

No opinion on `@` as a separator but I like the idea of a separator here.

> Is it OK that we are using private APIs of the Scrapy downloader? Should we try to make upstream to expose them as public APIs, be it before or after this change?

AFAICS these are `Downloader._get_slot_key()` and `Downloader._get_slot()`? I think it's fine to promote them to public. I've also found two places mentioning subclassing the download to override `_get_slot()`: https://github.com/scrapy/scrapy/issues/1458 and https://github.com/scrapy/scrapy/issues/5511

scrapy_zyte_api/_downloader_middleware.py
```diff
@@ -0,0 +1,32 @@
...
+        if self._param_parser.parse(request) is None:
+            return
+
+        downloader = self._crawler.engine.downloader
+        slot_id = downloader._get_slot_key(request, spider)
+        if not slot_id.startswith(self._slot_prefix):
+            slot_id = f"{self._slot_prefix}{slot_id}"
+            request.meta["download_slot"] = slot_id
+        if slot_id in self._handled_slots:
+            return
```
kmike, line 29: is it ok to set delay to 0 only once per slot? It seems AutoThrottle adjusts slot delays after each response; won't this 0 be overwritten by AutoThrottle?

Gallaecio: I think it should be OK at the moment.

> The variable part of AutoThrottle, i.e. the adjustment of the download delay based on response time, does not actually affect Zyte API requests. Our custom handler.download_request method does not report the download_latency in response.meta, causing [this early return in AutoThrottle](https://github.com/scrapy/scrapy/blob/master/scrapy/extensions/throttle.py#L43-L44).

If we plan to eventually have our download handler set the latency, then yes, we might want to do this every time. We might have a bigger problem, though, because I am not sure of the order in which extensions and middelwares execute, and depending on that order it may happen that the extension changes the slot delay after we set it to 0 but before it is taken into account by the Scrapy core.

kmike: I'm not sure I follow @Gallaecio, could you please explain it again? :) My understanding:

1. The whole issue is that AutoThrottle sets delay to slots, and we don't want any delays set by AuthThrottle, as it's handled server-side.
2. It means that for these custom slots we always want to use 0 delay.
3. AutoThrottle updates the slot delay after each response received for the slot.
4. We set delay to 0 only once per slot. 
5. It means that the 0 is going to be overwritten by AutoThrottle when another request to the same domain is made, so the (1) is not solved, as (2) is not happening.

Gallaecio: > 3. AutoThrottle updates the slot delay after each response received for the slot.

Specifically, AutoThrottle changes the initial delay of every slot, _and_ updates that delay based on response latency.

To alter the initial delay of every slot, [AutoThrottle sets `Spider.download_delay` to the maximum value of `Spider.download_delay`, `DOWNLOAD_DELAY`, and `AUTOTHROTTLE_START_DELAY`](https://github.com/scrapy/scrapy/blob/0417e3ede2e9388460ebe6e489de5ffd08580a03/scrapy/extensions/throttle.py#L28), which by default is the value of the later, [5 seconds](https://github.com/scrapy/scrapy/blob/0417e3ede2e9388460ebe6e489de5ffd08580a03/scrapy/settings/default_settings.py#L27). When a slot is created, [`Spider.download_delay` is set as the slot delay if defined](https://github.com/scrapy/scrapy/blob/06f3d12c1208c380f9f1a16cb36ba2dfa3c244c5/scrapy/core/downloader/__init__.py#L61).

We address this with (4).

To update that delay based on response latency, AutoThrottle relies on the `download_latency` Request.meta key, which download handlers set[¹](https://github.com/scrapy/scrapy/blob/9e265a2c1f6bccb551e8292785e09462594b8402/scrapy/core/downloader/webclient.py#L112)[²](https://github.com/scrapy/scrapy/blob/04a1a4b7303632268ece2b749e3326790d0c9f8c/scrapy/core/downloader/handlers/http11.py#L350)[³](https://github.com/scrapy/scrapy/blob/04a1a4b7303632268ece2b749e3326790d0c9f8c/scrapy/core/downloader/handlers/http2.py#L107). The custom downloader handler of scrapy-zyte-api, however, does not set this Request.meta key (which could be considered a bug) on requests that go through Zyte API, [preventing AutoThrottle from changing the slot delay based on response latency](https://github.com/scrapy/scrapy/blob/0417e3ede2e9388460ebe6e489de5ffd08580a03/scrapy/extensions/throttle.py#L42-L44).

So, as long as our custom download handler continues _not_ to record the download latency for Zyte API requests, (4) is enough to solve (1) and (2), because (3)/(5) does not happen for slots specific to Zyte API.

The only problem would be is a user manually assigned a request without Zyte API parameters to a Zyte API slot, as the latency of that request would affect the slot delay, because it would go through the default download handler that does set the request download latency. However, I think it is OK not to prepare for this scenario. This is the kind of scenario where users should be told “Do not do this unless you know what you are doing”.

kmike: Aha, I missed `download_latency` part, thanks for explaining it! 
Last question: why do we need `_handled_slots`?

Gallaecio: We don’t. It is meant as an optimization to avoid repeated calls to the code below, but that code is probably not expensive enough to justify the use of `_handled_slots`.

On the other hand, it could be argued that it helps readability by making it more obvious that we are only resetting the slot delay once, and not on every request processing.

I have no preference on whether to remove it or keep it, shall I remove it?

kmike: I think it is better to remove it: 

* it makes the code a bit more complex (though that's debatable),
* it may increase memory usage,
* I'm not sure it provides a speedup, and 
* we may face other gotchas with it - e.g. there is slot gc in downloader, so a slot can be absent in downloader, but present in handled_slots.

README.rst
```diff
@@ -70,6 +75,9 @@ For example, in the ``settings.py`` file of your Scrapy project:
         "http": "scrapy_zyte_api.ScrapyZyteAPIDownloadHandler",
         "https": "scrapy_zyte_api.ScrapyZyteAPIDownloadHandler",
     }
+    DOWNLOADER_MIDDLEWARES = {
+        "scrapy_zyte_api.ScrapyZyteAPIDownloaderMiddleware": 1000,
```
kmike, line 79: Not to solve here, by "Scrapy" in the middleware name looks unnecessary. "ScrapyZyteAPI" also may be unnecessary: `scrapy_zyte_api.DownloaderMiddleware`. I think we shouldn't change it in this PR, to maintain consistency with other options.

kmike:
> Should we rename the new downloader middleware after what it does (name suggestions?), or keep this generic name and try to concentrate any downloader middleware code that we may need in the future into this downloader middleware, instead of creating new downloader middlewares?

I think generic name is better. It may allow us to put there other features needed for Zyte API to work properly, without users adding more things to settings.py. See also https://github.com/scrapy-plugins/scrapy-zyte-api/pull/59/files#r1035262742.

> Is it OK that the whole Zyte API parameter parsing is done just to determine whether or not Zyte API is used, or should we use a simpler, cheaper implementation for this new purpose?
> Is it OK that we are calling the Zyte API parameter parsing logic 3 times per request now? Should we cache those calls somehow? (I assume weak references to requests are not reliable here, as meta can change, and I am not sure what the best alternative would be)

No strong opinion on this. I guess we can always optimize it later.

> Is zyte-api@ the right prefix for slot IDs? zyte-api@example.com looks cool to me, but I wonder if it can be misleading, or if @ can be problematic.

Prefix looks good.

> Is it OK that we are using private APIs of the Scrapy downloader? Should we try to make upstream to expose them as public APIs, be it before or after this change?

I think a public API for slots can be good, but I'm not sure if current API is good as a public API. Maybe it is. For now, it's ok to use private APIs of Scrapy downloader here.

> Should I refactor _params.py to turn its functions into methods of its new class, to minimize parameter passing?

If you have an idea for a cleanup, go for it :) It's not a requirement for merging this PR.

kmike (approved):
👍

kmike:
It seems nothing prevents us from merging this PR :) Thanks @Gallaecio!

PR #60, opened 2022-11-29
Reuse AsyncClient where possible

### LINKED ISSUE #58: Inconsistencies in stats when performing additional requests inside Page Objects
ARs inside POs currently work when **scrapy-zyte-api** sets `ZYTE_API_TRANSPARENT_MODE=True` in its settings.

However, this causes some inconsistencies in Scrapy stats:

```
{
 'downloader/request_count': 3,
 'downloader/request_method_count/GET': 3,
 'downloader/response_count': 3,
 'downloader/response_status_count/200': 3,
 
 'scrapy-zyte-api/429': 0,
 'scrapy-zyte-api/attempts': 1,
 'scrapy-zyte-api/error_ratio': 0.0,
 'scrapy-zyte-api/errors': 0,
 'scrapy-zyte-api/fatal_errors': 0,
 'scrapy-zyte-api/processed': 1,
 'scrapy-zyte-api/status_codes/200': 1,
 'scrapy-zyte-api/success': 1,
 'scrapy-zyte-api/success_ratio': 1.0,
 'scrapy-zyte-api/throttle_ratio': 0.0,
}
```

The `scrapy-zyte-api/*` stats should also reflect the ARs performed inside POs.

Gallaecio:
Fixes #58 by keeping a cache of Zyte API client objects, and reusing it.

Known issue: the cache does not take into account `os.environ["ZYTE_API_KEY"]`. It would be weird for that value to change mid-execution, so I think it is OK to ignore this issue. Alternatively, we could make it part of the key and client parameters, rather than passing None if the setting is not defined and letting zyte-api pick it from the environment variable instead.

Alternatives I have considered:
- Going for a singleton for the client object, rather than a cache, which should be fine as long as we do not want to allow the download handler class to be instantiated with a different crawler object during the same execution.
- Making the whole download handler class a wrapper, and having all wrappers reuse a singleton implementation of the whole download handler. Discarded as a rather complex solution for the scope of #58. It actually seems best to me assuming there is no issue in having the same object handle different protocols, and I wonder if we should do this in Scrapy’s download handler itself.
- Sharing the stats object instead. To avoid using private API, it requires changes in zyte-api to allow passing stats by parameter to the `__init__` method of the client class. But I see no point in this approach if using the same client also works, since the latter is easier to implement and avoids additional run-time duplicity.
- Having the stat update method somehow aggregate the stats from all running clients. It seems rather complicated compared to any other solution.

Note: I have only addressed mypy issues that seemed related to these specific changes, anything else should hopefully be addressed by #59.

### DIFF
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, Generator, Optional, Tuple, Union
+from typing import Generator, Optional, Union
 
 from scrapy import Spider
 from scrapy.core.downloader.handlers.http import HTTPDownloadHandler
@@ -29,9 +29,6 @@ def _load_retry_policy(settings):
     return policy
 
 
-_CLIENT_CACHE: Dict[Tuple[Optional[str], str, int], AsyncClient] = {}
-
-
 class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
     def __init__(
         self, settings: Settings, crawler: Crawler, client: AsyncClient = None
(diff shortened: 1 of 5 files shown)


scrapy_zyte_api/handler.py
```diff
@@ -34,6 +34,9 @@ def _load_retry_policy(settings):
     return policy
 
 
+_CLIENT_CACHE: Dict[Tuple[Optional[str], str, int], AsyncClient] = {}
```
BurnzZ, line 37: I think it'd be worth noting the issue #58 here as a comment, in case it gets refactored/removed later.

scrapy_zyte_api/handler.py
```diff
@@ -42,24 +45,33 @@ def __init__(
...
+                        api_url=client_key[1],
+                        n_conn=client_key[2],
+                    )
+                except NoApiKey:
+                    logger.warning(
+                        "'ZYTE_API_KEY' must be set in the spider settings or env var "
+                        "in order for ScrapyZyteAPIDownloadHandler to work."
+                    )
+                    raise NotConfigured
+                _CLIENT_CACHE[client_key] = client
```
BurnzZ, line 74: optional: this client creation code-block is getting large, we could put it in another method.

BurnzZ (approved):
Great find! 🙌

scrapy_zyte_api/handler.py
kmike, line 74: One more alternative: store the client object as an attribute of either `crawler` or `spider`. Basically, we need some storage which is shared between all components in a crawl. If we use spider, it would also allow users to define custom AsyncClient instance as a spider attribute.

kmike: If we use cache with settings as keys, then stats are going to be shared between spiders running in the same process, if the options are the same.

tests/test_handler.py
```diff
@@ -284,3 +284,18 @@ async def test_stats(mockserver):
...
+    crawler = get_crawler(settings_dict=SETTINGS)
+    handler1 = ScrapyZyteAPIDownloadHandler(
+        settings=crawler.settings,
+        crawler=crawler,
+    )
+    handler2 = ScrapyZyteAPIDownloadHandler(
+        settings=crawler.settings,
+        crawler=crawler,
+    )
+    assert handler1._client == handler2._client
```
kmike, line 301:
```suggestion
    assert handler1._client is handler2._client
```

kmike (approved):
👍

PR #61, opened 2022-12-09
Prepare for 0.7.0 release

BurnzZ:
# PRs
- https://github.com/scrapy-plugins/scrapy-zyte-api/pull/56
- https://github.com/scrapy-plugins/scrapy-zyte-api/pull/59

# Checklist
- [x] Update release date in CHANGES.rst
- [ ] Tag `0.7.0` after merging

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,46 @@
 Changes
 =======
 
+0.7.0 (TBD)
+-----------
+
+To upgrade, you need to set the following in your Scrapy project settings:
+
+.. code-block:: python
+
+  DOWNLOADER_MIDDLEWARES = {
+      "scrapy_zyte_api.ScrapyZyteAPIDownloaderMiddleware": 1000,
+  }
+  # only applicable for Scrapy 2.7+
+  REQUEST_FINGERPRINTER_CLASS = "scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter"
+
+* Fixes the issue where scrapy-zyte-api is slow when Scrapy Cloud has Autothrottle
+  Addon enabled. The new ``ScrapyZyteAPIDownloaderMiddleware`` fixes this.
+
+* It now supports Scrapy 2.7's new ``REQUEST_FINGERPRINTER_CLASS`` which ensures
+  that Zyte API requests are properly fingerprinted. This addresses the issue
+  where Scrapy marks POST requests as duplicate if it points to the same URL
+  despite having different request bodies. As a workaround, users were marking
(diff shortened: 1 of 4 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,46 @@
...
+  }
+  # only applicable for Scrapy 2.7+
+  REQUEST_FINGERPRINTER_CLASS = "scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter"
+
+* Fixes the issue where scrapy-zyte-api is slow when Scrapy Cloud has Autothrottle
+  Addon enabled. The new ``ScrapyZyteAPIDownloaderMiddleware`` fixes this.
+
+* It now supports Scrapy 2.7's new ``REQUEST_FINGERPRINTER_CLASS`` which ensures
+  that Zyte API requests are properly fingerprinted. This addresses the issue
+  where Scrapy marks POST requests as duplicate if it points to the same URL
```
wRAR, line 22: "if they point" maybe?

wRAR (approved):

tox.ini
```diff
@@ -28,43 +28,43 @@ deps =
 
 
 # Earliest supported Scrapy version.
-[testenv:pinned-scrapy-2.0]
```
BurnzZ, line 31: The new update on `tox >= 4.0` had a lot of backward incompatible changes.

It's now strict on the `testenv:<env_name>` format which renders this old line to have this error:

```
pinned-26-scrapy: failed with env name pinned-scrapy-2.0 conflicting with base python python3.7
pinned-scrapy-2.0: FAIL code 1 (0.01 seconds)
evaluation failed :( (1.24 seconds)
```

Setting `ignore_basepython_conflict=True` is supposed to fix this but it doesn't work. Seems like they're still releasing a lot of regression fixes after the `4.0` release.

They're also quite aggressive on the name inference. If they see any two digits in the env name, they'd treat it as a Python version. So other workarounds like `20`, `2-0`, etc doesn't work.

CHANGES.rst
```diff
@@ -1,6 +1,46 @@
 Changes
 =======
 
+0.7.0 (TBD)
+-----------
+
+To upgrade, you need to set the following in your Scrapy project settings:
```
Gallaecio, line 7: - It is not technically necessary
- These Scrapy settings can be defined at the spider level

```suggestion
When upgrading, you should set the following in your Scrapy settings:
```

Gallaecio (approved):

kmike (approved):

PR #62, opened 2022-12-11
Readme fix zyte_api_automap code snippet

felipeboffnunes:
I may be wrong, I don't think so but who knows.

I tried zyte_api_automap as the snippet of the Readme said to do. It broke.
When I added directly to `Request.meta`, as the description says, it worked and the `Response` got the `raw_api_response` field as expected.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -243,9 +243,7 @@ For example:
             yield scrapy.Request(
                 url="https://quotes.toscrape.com/",
                 meta={
-                    "zyte_api": {
-                        "zyte_api_automap": True,
-                    }
+                    "zyte_api_automap": True,
                 },
             )
 
--- a/README.rst
+++ b/README.rst
@@ -365,9 +365,7 @@ parameters are chosen as follows by default:
             Request(
                 url="https://toscrape.com/img/zyte.png",
                 meta={
-                    "zyte_api": {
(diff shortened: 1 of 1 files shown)


BurnzZ (changes requested):
Good catch! Yeah I think this is the expected behavior, as per the following:
- https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/_params.py#L362
- https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/_params.py#L416

If so, I think we also need to fix line 371 and 391 of the README as well.

Another route is to formally support `zyte_api_automap` inside `zyte_api` as well, which has some of its merits as well. @Gallaecio what do you think?

Gallaecio:
Supporting `zyte_api_automap` inside `zyte_api` was never the intention, it was a mistake on my part. If there are more cases, let’s fix them all with this pull request.

BurnzZ (approved):
Thanks @felipeboffnunes !

@Gallaecio to merge if these are all the changes we need.

felipeboffnunes:
For now, I think that is the only error I have found from all the things I needed from the plugin. I can test later on the other features but no need to keep the PR open, I can open a new one if I find anything else.

Gallaecio (approved):

PR #64, opened 2023-01-07
Allow logging request parameters

### LINKED ISSUE #48: Log the actual request sent to Zyte API
When requests go through Zyte API, it may be useful for debugging purposes to log exactly which parameters are sent to Zyte API (i.e. the JSON API request body).

This should become specially useful once we merge #41.

Gallaecio:
Resolves #48.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -733,4 +733,5 @@ For example::
    Sending Zyte API extract request: {"url": "https://example.com", "httpResponseBody": true}
 
 The ``ZYTE_API_LOG_REQUESTS_TRUNCATE``, 64 by default, determines the maximum
-length of any string value in the logged JSON object, excluding object keys.
+length of any string value in the logged JSON object, excluding object keys. To
+disable truncation, set it to 0.
--- a/README.rst
+++ b/README.rst
@@ -719,3 +719,18 @@ prevent the duplicate filter of Scrapy to filter any of them out. For example:
 
 Note, however, that for other Scrapy components, like the HTTP cache
 extensions, these 2 requests would still be considered identical.
+
+
+Logging request parameters
+==========================
+
+Set the ``ZYTE_API_LOG_REQUESTS`` setting to ``True`` to enable the logging of
+debug messages that indicate the JSON object sent on every extract request to
+Zyte API.
+
+For example::
+
+   Sending Zyte API extract request: {"url": "https://example.com", "httpResponseBody": true}
+
+The ``ZYTE_API_LOG_REQUESTS_TRUNCATE``, 64 by default, determines the maximum
+length of any string value in the logged JSON object, excluding object keys.
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -85,6 +85,12 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
         self._session = create_session(connection_pool_size=self._client.n_conn)
         self._must_log_request = settings.getbool("ZYTE_API_LOG_REQUESTS", False)
         self._truncate_limit = settings.getint("ZYTE_API_LOG_REQUESTS_TRUNCATE", 64)
+        if self._truncate_limit < 0:
+            raise ValueError(
+                f"The value of the ZYTE_API_LOG_REQUESTS_TRUNCATE setting "
+                f"({self._truncate_limit}) is invalid. It must be 0 or a "
+                f"positive integer."
+            )
 
     def download_request(self, request: Request, spider: Spider) -> Deferred:
         api_params = self._param_parser.parse(request)
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -26,8 +26,7 @@ logger = logging.getLogger(__name__)
 
 def _truncate(obj, limit):
     if isinstance(obj, dict):
-        for key in list(obj):
-            value = obj[key]
+        for key, value in obj.items():
             if isinstance(value, str):
                 if len(value) > limit:
                     obj[key] = value[: limit - 1] + "…"
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -1,4 +1,6 @@
+import json
 import logging
+from copy import deepcopy
 from typing import Generator, Optional, Union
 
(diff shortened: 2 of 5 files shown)


scrapy_zyte_api/handler.py
```diff
@@ -22,6 +24,24 @@
...
 
 
+def _truncate(obj, limit):
+    if isinstance(obj, dict):
+        for key in list(obj):
+            value = obj[key]
+            if isinstance(value, str):
+                if len(value) > limit:
+                    obj[key] = value[: limit - 1] + "…"
+            elif isinstance(value, (list, dict)):
```
BurnzZ, line 34: Do we also allow and process any other API request parameters that are iterables like `tuples` and `sets`?

README.rst
```diff
@@ -719,3 +719,18 @@ prevent the duplicate filter of Scrapy to filter any of them out. For example:
...
+Set the ``ZYTE_API_LOG_REQUESTS`` setting to ``True`` to enable the logging of
+debug messages that indicate the JSON object sent on every extract request to
+Zyte API.
+
+For example::
+
+   Sending Zyte API extract request: {"url": "https://example.com", "httpResponseBody": true}
+
+The ``ZYTE_API_LOG_REQUESTS_TRUNCATE``, 64 by default, determines the maximum
+length of any string value in the logged JSON object, excluding object keys.
```
BurnzZ, line 736: Let's also mention what happens when the value is `0` or `< 0`.

scrapy_zyte_api/handler.py
```diff
@@ -22,6 +24,24 @@
 logger = logging.getLogger(__name__)
 
 
+def _truncate(obj, limit):
+    if isinstance(obj, dict):
+        for key in list(obj):
+            value = obj[key]
```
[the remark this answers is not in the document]
Gallaecio: I was afraid of issues with modifying the dict while iterating it. But [I see the specific case of modifying existing indexes is not problematic](https://stackoverflow.com/a/6777632/939364), so :+1:

scrapy_zyte_api/handler.py
Gallaecio: Since it is a JSON API, and we use `json` from the standard library to parse it, I think it is safe to assume we will only have to handle lists and dicts here.

scrapy_zyte_api/handler.py
```diff
@@ -22,6 +24,24 @@
...
+            if isinstance(value, str):
+                if len(value) > limit:
+                    obj[key] = value[: limit - 1] + "…"
```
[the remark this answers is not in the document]
Gallaecio: Done, although I am not sure if the implementation I went with is OK, or if you would prefer e.g. the `obj[index] = _truncate_str(value, limit)` approach. No strong opinion from my end.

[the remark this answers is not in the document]
kmike: No strong opinion; you implementation looks fine 👍

tests/test_handler.py
```diff
@@ -284,3 +287,130 @@ async def test_stats(mockserver):
...
+
+@ensureDeferred
+@pytest.mark.skipif(sys.version_info < (3, 8), reason="unittest.mock.AsyncMock")
+@pytest.mark.parametrize(
+    "settings,short_str,long_str,truncated_str",
+    [
+        ({}, "a" * 64, "a" * 65, "a" * 63 + "…"),
+        ({"ZYTE_API_LOG_REQUESTS_TRUNCATE": 0}, "a" * 64, "a" * 65, "a" * 65),
+        ({"ZYTE_API_LOG_REQUESTS_TRUNCATE": 1}, "a", "aa", "…"),
+        ({"ZYTE_API_LOG_REQUESTS_TRUNCATE": 2}, "aa", "aaa", "a…"),
```
Gallaecio, line 325: In the implementation, I went with an Unicode ellipsis (…) thinking Windows now supports UTF-8 terminal output. But re-reading about it, it seems users actually need to configure things to get UTF-8 terminal output, it is not enabled by default.

The problem with the ASCII alternative (...) is that it is 3-characters long. Shall we set a minimum value for truncation of 3 instead of 1, and disallow 1 and 2 as values for the setting?

BurnzZ: >  it seems users actually need to configure things to get UTF-8 terminal output, it is not enabled by default.
Could you send a link to a doc/article regarding this?

I'm testing it in my Windows machine and "…" currently works without any problems.

tests/test_handler.py
```diff
@@ -299,3 +302,130 @@ def test_single_client():
...
+        actual_api_params = handler._client.request_raw.call_args[0][0]
+        del actual_api_params["url"]
+        assert actual_api_params == expected_api_params
+
+
+def test_log_request_truncate_negative():
+    settings: Dict[str, Any] = {
+        **SETTINGS,
+        "ZYTE_API_LOG_REQUESTS": True,
+        "ZYTE_API_LOG_REQUESTS_TRUNCATE": -1,
```
BurnzZ, line 423: Perhaps we should also test if `ValueError` is still raised when `ZYTE_API_LOG_REQUESTS=False`.

scrapy_zyte_api/handler.py
```diff
@@ -159,6 +191,19 @@ async def _download_request(
 
         return _process_response(api_response, request)
 
+    def _log_request(self, params):
+        if not self._must_log_request:
+            return
+        params = self._truncate_params(params)
+        logger.debug(f"Sending Zyte API extract request: {json.dumps(params)}")
```
BurnzZ, line 198: Should we also log the API endpoint that we're sending the request through? I.e. https://api.zyte.com/v1/extract.

My concern is that users might be interested if they're using a newer v2 or v3 endpoint later on. Or perhaps their own dedicated enterprise endpoint for that matter.

We could also choose to log this API endpoint once when the handler is created.

Gallaecio: > We could also choose to log this API endpoint once when the handler is created.

Yes, I think this is a good point but it may be best to delay it until we support alternative endpoints. And, even then, we might want to log it once at the beginning of the logs, to keep request-specific logs as short as possible, unless we support using multiple different endpoints on the same spider.

BurnzZ (approved):

kmike:
👍 👍

PR #69, opened 2023-02-25
Cookie support for automap

Gallaecio:
To-do:
- [x] Handle `TODO` comments.
- [x] Separate cookie parsing and formatting into a separate module that could eventually serve as a starting point for https://github.com/scrapy/scrapy/issues/5431
- [x] Agree on the best way to handle max cookies.
- [ ] Should we provide a workaround for https://github.com/scrapy/scrapy/issues/5841? Push for an upstream fix? Let things be for now?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -362,8 +362,9 @@ parameters are chosen as follows by default:
 
 -   ``Request.body`` becomes ``httpRequestBody``.
 
--   If the COOKIES_ENABLED_ Scrapy setting is ``True`` (default), and unless
-    request metadata sets dont_merge_cookies_ to ``True`` as well:
+-   If the ``ZYTE_API_EXPERIMENTAL_COOKIES_ENABLED`` Scrapy setting is
+    ``True``, the COOKIES_ENABLED_ Scrapy setting is ``True`` (default), and
+    provided request metadata does not set dont_merge_cookies_ to ``True``:
 
     .. _COOKIES_ENABLED: https://docs.scrapy.org/en/latest/topics/downloader-middleware.html#std-setting-COOKIES_ENABLED
     .. _dont_merge_cookies: https://docs.scrapy.org/en/latest/topics/request-response.html#std-reqmeta-dont_merge_cookies
@@ -381,11 +382,12 @@ parameters are chosen as follows by default:
         during browser rendering).
 
         If the cookies to be set exceed the limit defined in the
-        ``ZYTE_API_MAX_COOKIES`` setting (20 by default), a warning is logged,
+        ``ZYTE_API_MAX_COOKIES`` setting (100 by default), a warning is logged,
         and no cookies are set at all for the target request. To silence this
         warning, set ``experimental.requestCookies`` manually, e.g. to an empty
-        dict. Alternatively, if Zyte API starts supporting more than 20 request
-        cookies, update the ``ZYTE_API_MAX_COOKIES`` setting accordingly.
+        dict. Alternatively, if Zyte API starts supporting more than 100
+        request cookies, update the ``ZYTE_API_MAX_COOKIES`` setting
+        accordingly.
 
         If you are using a custom downloader middleware to handle request
         cookiejars, you can point the ``ZYTE_API_COOKIE_MIDDLEWARE`` setting to
--- a/README.rst
+++ b/README.rst
@@ -383,11 +383,11 @@ parameters are chosen as follows by default:
 
         If the cookies to be set exceed the limit defined in the
         ``ZYTE_API_MAX_COOKIES`` setting (100 by default), a warning is logged,
-        and no cookies are set at all for the target request. To silence this
-        warning, set ``experimental.requestCookies`` manually, e.g. to an empty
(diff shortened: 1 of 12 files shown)


README.rst
```diff
@@ -351,6 +352,48 @@ mode (see **Using transparent mode** above) or for a specific request (see
...
 
+-   ``Request.url`` becomes ``url``, same as in requests with manually-defined
+    parameters.
+
+-   If ``Request.method`` is something other than ``"GET"``, it becomes
+    ``httpRequestMethod``.
+
+-   ``Request.headers`` become ``customHttpRequestHeaders``.
+
+-   ``Request.body`` becomes ``httpRequestBody``.
```
Gallaecio, line 363: Moved at the top to get the short bullet points out of the way earlier and have the more verbose bullet points towards the end.

README.rst
```diff
@@ -351,6 +352,48 @@ mode (see **Using transparent mode** above) or for a specific request (see
...
+        domain. This is because Zyte API requests may involve requests to
+        different domains (e.g. when following cross-domain redirects, or
+        during browser rendering).
+
+        If the cookies to be set exceed the limit defined in the
+        ``ZYTE_API_MAX_COOKIES`` setting (20 by default), a warning is logged,
+        and no cookies are set at all for the target request. To silence this
+        warning, set ``experimental.requestCookies`` manually, e.g. to an empty
+        dict. Alternatively, if Zyte API starts supporting more than 20 request
+        cookies, update the ``ZYTE_API_MAX_COOKIES`` setting accordingly.
```
Gallaecio, line 388: I am really unsure about what to do here. At one point I decided to remove this check altogether and do like we do with mixing httpResponseBody and browserHtml, i.e. let users send the wrong thing and let Zyte API report the issue.

However, mixing those 2 outputs never happens with plain automap, whereas the cookie limit can easily happen.

But even if we agree on keeping some form of handling on the client side, I am not sure what the best handling is. The proposal here logs a warning and sets no cookies at all. But that can often mean breaking the request. Setting 20 cookies could be just as bad. I wonder if it would not be better to directly raise an exception and not send the request at all.

README.rst
```diff
@@ -351,6 +352,48 @@ mode (see **Using transparent mode** above) or for a specific request (see
...
+
+        If you are using a custom downloader middleware to handle request
+        cookiejars, you can point the ``ZYTE_API_COOKIE_MIDDLEWARE`` setting to
+        its import path to make scrapy-zyte-api work with it. The downloader
+        middleware is expected to have a ``jars`` property with the same
+        signature as in the built-in Scrapy downloader middleware for cookie
+        handling.
```
Gallaecio, line 395: This is on the off chance that someone needs this level of flexibility. Going forward, though, I think we should consider centralizing the cookie jars storage on the crawler in upstream Scrapy, similar to how we did with request fingerprinting.

README.rst
kmike: If cookies are dropped, there is a chance request still may work. If we raise an exception, it seems that there is no workaround for the user. So, it seems it'd be better to have a warning, not an exception. What do you think?

The limit is going to be increased to 100 soon; it's not a complete solution, but, hopefully, it'd help a bit.

README.rst
```diff
@@ -351,6 +352,48 @@ mode (see **Using transparent mode** above) or for a specific request (see
...
+
+-   If ``Request.method`` is something other than ``"GET"``, it becomes
+    ``httpRequestMethod``.
+
+-   ``Request.headers`` become ``customHttpRequestHeaders``.
+
+-   ``Request.body`` becomes ``httpRequestBody``.
+
+-   If the COOKIES_ENABLED_ Scrapy setting is ``True`` (default), and unless
+    request metadata sets dont_merge_cookies_ to ``True`` as well:
```
kmike, line 366: I wonder if cookie handling should make it opt-in. According to Zyte API docs, this is an experimental feature which can be dropped or changed at any time without releasing a new API version. Not sure though, the cons is that users won't get cookie handling by default.

Gallaecio: > The cons is that users won't get cookie handling by default.

We can warn about it the first time we get a requests where cookie would have been set, so I think it is reasonable to make it opt-in. On it…

README.rst
kmike: Hm, I was thinking that the warning is going to be issued, but the cookies which are under the limit are going to be sent. What's the advantage of dropping them all?

Gallaecio: No user-facing advantage.

If sending some cookies, though, I think we should discuss:
- Prioritization: should we prioritize some cookies over other, e.g. based on the target domain or how long the cookies have been in the cookie jar (e.g. the more recent, the higher the priority)?
- Communication: when dropping cookies, should we indicate, e.g. through a log message, which cookies are sent or which cookies are dropped? Or at least how many?

kmike:
Thanks @Gallaecio! I've checked all the changes, and can't spot any issue 👍  This is a complex feature, so we both might be missing something, but it'd be found in a real-world usage. So, let's proceed. Great work!

PR #70, opened 2023-03-09
Support non-str slots

Gallaecio:


Gallaecio:
I am unsure whether we should implement this, or let it be, so that users using e.g. `int` are forced to rethink how they are using slots.

kmike:
I definitely expected slots to be strings; we should probably document it as a part of https://github.com/scrapy/scrapy/issues/3585.

The workaround here looks simple enough though, so no pushback for merging.

wRAR (approved):

Issue #67, opened 2023-02-13
Scrapy's open_in_browser never uses .html extension for Zyte API responses


kmike:
Scrapy's open_in_browser uses response type to pick the extension of the saved file.

But see https://github.com/scrapy-plugins/scrapy-zyte-api/blob/27770881e739190655239e00644886851310f2f5/scrapy_zyte_api/responses.py#L73 - the response is a TextResponse subclass, and never a HtmlResponse instance. So, open_in_browser never uses a correct .html extension for Zyte API responses.

PR #71, opened 2023-03-20
Changed TextResponse for HtmlResponse

manu189:
Fixes https://github.com/scrapy-plugins/scrapy-zyte-api/issues/67

### DIFF
--- a/scrapy_zyte_api/responses.py
+++ b/scrapy_zyte_api/responses.py
@@ -2,7 +2,7 @@ from base64 import b64decode
 from typing import Dict, List, Optional, Tuple, Union
 
 from scrapy import Request
-from scrapy.http import Response, TextResponse
+from scrapy.http import Response, TextResponse, HtmlResponse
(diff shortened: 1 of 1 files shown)


kmike (approved):
Thanks @manu189! I think the ideal fix would be to have some logic similar to what's used in Scrapy, to use TextResponse base class if the response is not html (e.g. robots.txt files). But I'm fine with merging this implementation, because:

1. It seems it's going to fix the most annoying issues
2. The "proper" approach could be much harder to implement, for insignificant benefits (unless we learn otherwise).

manufacio:
HI @kmike  being that HtmlResponse is basically equal to TextResponse, wouldn't it be a proper fix? 
image

kmike:
hey! I was also thinking they’re equal, but as it turns out, it behaves differently e.g. in scrapy’s open_in_browser. It should be rare though that there is any practical difference, but there could be code which works in a different way, depending on the response type.

PR #73, opened 2023-03-30
add anchor links to readme

martinhohoff:


Gallaecio (approved):
I‘m surprised that it works this way, without defining target anchors, but [it works](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/93224ab1975618ea3b66d5c0960201a72125e74e/README.rst#using-transparent-mode), so +1.

It is indeed an significant improvement until we decide to switch to Sphinx.

martinhohoff:
> I‘m surprised that it works this way, without defining target anchors, but [it works](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/93224ab1975618ea3b66d5c0960201a72125e74e/README.rst#using-transparent-mode), so +1.
> 
> It is indeed an significant improvement until we decide to switch to Sphinx.

It was news to me as well. Easy syntax and usage.

PR #88, opened 2023-05-26
Get the Scrapy Cloud job ID from the SHUB_JOBKEY environment variable

Gallaecio:
Based on https://shub.readthedocs.io/en/stable/custom-images-contract.html#shub-jobkey.

Verified manually on a Scrapy Cloud project.

### DIFF
--- a/scrapy_zyte_api/_params.py
+++ b/scrapy_zyte_api/_params.py
@@ -1,6 +1,7 @@
 from base64 import b64decode, b64encode
 from copy import copy
 from logging import getLogger
+from os import environ
 from typing import Any, Dict, List, Mapping, Optional, Set
 from warnings import warn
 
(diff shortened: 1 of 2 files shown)


scrapy_zyte_api/_params.py
```diff
@@ -569,7 +570,7 @@ def __init__(self, crawler, cookies_enabled=None):
         self._automap_params = _load_default_params(settings, "ZYTE_API_AUTOMAP_PARAMS")
         self._browser_headers = _load_browser_headers(settings)
         self._default_params = _load_default_params(settings, "ZYTE_API_DEFAULT_PARAMS")
-        self._job_id = settings.get("JOB")
+        self._job_id = environ.get("SHUB_JOBKEY", None)
```
kmike, line 573: Is it the case that JOB doesn't work, or is it a clean up to use a documented env variable?

Gallaecio: The JOB setting does not work. I am not sure it ever did.

kmike (commented):
Thanks @Gallaecio! +1 to merge after fixing tests

scrapy_zyte_api/_params.py
pawelmhm: I've seen job IDs in several log entries, I didn't investigate how they found their ways to logs, were projects using this plugin or something else. I know about some projects where job key is actually set in settings. 

It is also possible some users went ahead and added JOB key in settings to track it. Perhaps we could keep this for backward compatibility?

Gallaecio: > Perhaps we could keep this old way for backward compatibility? It is possible some users are using this plugin outside scrapy cloud and perhaps they rely on this key for something

I would rather mention the backward incompatibility in the changelog, and indicate that if they want something like that they can set the environment variable locally or use `echoData` instead.

Issue #91, opened 2023-06-12
ZyteApiProvider could make an unneeded API request 


kmike:
In the example below ZyteApiProvide makes 2 API requests instead of 1:

```py
@handle_urls("example.com")
@attrs.define
class MyPage(ItemPage[MyItem]):
    html: BrowserHtml
    # ...

class MySpider(scrapy.Spider):
    # ...
    def parse(self, response: DummyResponse, product: Product, my_item: MyItem):
        # ...
```

Gallaecio:
Findings so far:
- https://github.com/scrapinghub/scrapy-poet/pull/151 won’t fix this.
- This issue seems to be caused by Zyte API provided classes being resolved at different stages. If you request both product and browser_response directly in the callback, a single request is sent. Otherwise, first Product is injected, then MyItem resolves to MyPage, then BrowserHtml is injected. I am not sure yet how to best solve that.

wRAR:
We also thought the solution may involve the caching feature in ItemProvider but didn't imvestigate further.

Gallaecio:
[Indeed](https://github.com/scrapy-plugins/scrapy-zyte-api/pull/94).

Gallaecio:
New finding: Switching `MyItem` to `MyPage` works, even if there is still some level of indirection. Could explain why https://github.com/scrapinghub/scrapy-poet/pull/153 works.

BurnzZ:
This issue also messes up how the `ZYTE_API_MAX_REQUESTS` setting is computed. PR: https://github.com/scrapy-plugins/scrapy-zyte-api/pull/114 

Example code:
```python
from web_poet import WebPage
from zyte_common_items import Product

handle_urls("example.com")
@attrs.define
class MyPage(WebPage[Product]):
    # ...

class MySpider(scrapy.Spider):
    # ...
    def parse(self, response: DummyResponse, product: Product):
        # ...
```

since `WebPage` contains a dependency for `response: web_poet.HttpResponse`, the dependency plan in **scrapy-poet**'s injector would be:

```
[
    (<class 'web_poet.page_inputs.http.HttpResponse'>, {}),                
    (<class 'zyte_common_items.items.Product'>, {}),             
    (<class 'my_project.page_objects.MyPage'>,                                           
     {'navigation_item': <class zyte_common_items.items.Product'>,         
      'response': <class 'web_poet.page_inputs.http.HttpResponse'>}),      
[...]

PR #94, opened 2023-06-15
Cache provider output per request

Gallaecio:
As a side effect, it allows to work around https://github.com/scrapy-plugins/scrapy-zyte-api/issues/91 by requesting indirect outputs directly on the spider callback as well.

### DIFF
--- a/scrapy_zyte_api/providers.py
+++ b/scrapy_zyte_api/providers.py
@@ -1,4 +1,5 @@
-from typing import Any, Callable, List, Sequence, Set
+from typing import Any, Callable, Dict, List, Sequence, Set, Type
+from weakref import WeakKeyDictionary
 
 from scrapy import Request
 from scrapy.crawler import Crawler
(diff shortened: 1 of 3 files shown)


wRAR (approved):

scrapy_zyte_api/providers.py
```diff
@@ -21,11 +22,29 @@ class ZyteApiProvider(PageObjectInputProvider):
 
     provided_classes = {BrowserResponse, BrowserHtml, Product}
 
+    def __init__(self, injector):
+        super().__init__(injector)
+        self._cached_instances: WeakKeyDictionary[Request, Dict] = WeakKeyDictionary()
```
kmike, line 27: We have a similar solution in ItemProvider. I wonder if it's possible to move this cache to Injector instead, so that there is a single cache which is shared across all providers.

The first step could be using the same cached_istances storage, and keeping all cache handling code as-is. But maybe it'd be possible to improve on this further, e.g. don't pass a class to to_provide if an instance is in the cache, and also automatically add results to the cache.

Gallaecio: I had this in my to-do, but I wonder if we need it now that ItemProvider is gone. I am not sure in which scenario the injector a provider would get the same Request +  deps now, given duplicate request filtering.

PR #95, opened 2023-06-20
Fix multiple API requests being sent by scrapy-poet where 1 should be enough

Gallaecio:
I have re-enabled the test about this, which fails as expected, but a test with a similar purpose added to https://github.com/scrapinghub/scrapy-poet/pull/153 passes, so I must be doing something wrong somewhere.


Gallaecio:
https://github.com/scrapy-plugins/scrapy-zyte-api/pull/156

PR #96, opened 2023-06-30
Add a Scrapy add-on

wRAR:
Post-release to-do:
- [ ] Update the [tutorial](https://docs.zyte.com/web-scraping/tutorial/index.html) and its project repo, and similar repos.


README.rst
```diff
@@ -42,7 +42,12 @@ Installation
 Configuration
 =============
 
-To enable this plugin:
+To enable this plugin, if your Scrapy version supports addons:
```
wRAR, line 45: Temporary wording

tests/test_handler.py
```diff
@@ -430,3 +430,12 @@ def test_log_request_truncate_negative(enabled):
             settings=None,
             crawler=crawler,
         )
+
+
+@ensureDeferred
+async def test_addon(mockserver):
```
wRAR, line 436: This test needs skipping on old Scrapy and I think this is the only place that check is needed.

README.rst
```diff
@@ -62,13 +67,24 @@ To enable this plugin:
...
     <https://docs.zyte.com/zyte-api/usage/general.html#authorization>`_ as
     either the ``ZYTE_API_KEY`` Scrapy setting or as an environment variable of
     the same name.
 
 For example, in the ``settings.py`` file of your Scrapy project:
 
+.. code-block:: python
+
+    ADDONS = {
+        "scrapy_zyte_api.ScrapyZyteAPIAddon": 1,
```
kmike, line 82: What do you think about using `scrapy_zyte_api.Addon` name?

scrapy_zyte_api/addon.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class ScrapyZyteAPIAddon:
+    def update_settings(self, settings: BaseSettings) -> None:
+        settings["DOWNLOAD_HANDLERS"][
+            "http"
+        ] = "scrapy_zyte_api.ScrapyZyteAPIDownloadHandler"
+        settings["DOWNLOAD_HANDLERS"][
+            "https"
+        ] = "scrapy_zyte_api.ScrapyZyteAPIDownloadHandler"
```
kmike, line 13: Should we add some checks for non-standard http/https handlers?

scrapy_zyte_api/addon.py
```diff
@@ -0,0 +1,24 @@
...
+        settings["DOWNLOADER_MIDDLEWARES"][ScrapyZyteAPIDownloaderMiddleware] = 1000
+        settings.set(
+            "REQUEST_FINGERPRINTER_CLASS",
+            "scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter",
+            "addon",
+        )
```
kmike, line 19: What's the behavior in case user sets their own request fingerprinter class? Can we do better than silently override it?

scrapy_zyte_api/addon.py
```diff
@@ -0,0 +1,24 @@
+from scrapy.settings import BaseSettings
+
+from scrapy_zyte_api import ScrapyZyteAPIDownloaderMiddleware
+
+
+class ScrapyZyteAPIAddon:
+    def update_settings(self, settings: BaseSettings) -> None:
```
kmike, line 7: I wonder if addon should set ZYTE_API_TRANSPARENT_MODE=True by default, as it's a recommended way to use the package.

kmike: In this case, we still should respect ZYTE_API_TRANSPARENT_MODE=False setting set by the user. 

Probably, we'd need a different default value for the option, not False, to be able to distinguish between these two cases, but maybe there is another way to solve it.

README.rst
wRAR: It's fine for the use case quoted here but I think it's worse when you want to import it, either for subclassing or to use the class in `ADDONS`.

scrapy_zyte_api/addon.py
wRAR: How about this?

I don't like this approach because it assumes the default values. We could import `scrapy.settings.default_settings` instead (hopefully no settings modify the mutable objects there), or for `DOWNLOAD_HANDLERS` just check its value without `getwithbase()`.

README.rst
Gallaecio: You can always rename on import, and the `ADDONS` setting would be a far more common use case, so I also think the shorter name may be worth it.

scrapy_zyte_api/addon.py
wRAR: Why do we need to distinguish them?

kmike: If user sets ZYTE_API_TRANSPARENT_MODE=False in their settings, we should respect that. I guess you can handle it by using settings priority directly, instead of having True/False/DEFAULT.

scrapy_zyte_api/addon.py
```diff
@@ -0,0 +1,51 @@
...
+            != "scrapy.utils.request.RequestFingerprinter"
+        ):
+            return "'REQUEST_FINGERPRINTER_CLASS'"
+        return None
+
+    def update_settings(self, settings: BaseSettings) -> None:
+        nondefault_msg = self._check_settings(settings)
+        if nondefault_msg:
+            raise ValueError(
+                f"The {nondefault_msg} setting is set to a custom value, refusing to override it"
```
kmike, line 32: On a high level, the logic for our handlers is this:

* if request is a Zyte API request, use python-zyte-api client
* otherwise, use the original implementation

Do you think it'd be possible to follow this logic in case of custom download handlers?

I think the same applies for request fingerprints; for non-Zyte-API requests ZYTE_API_FALLBACK_REQUEST_FINGERPRINTER_CLASS is used, which is the default one by default.

scrapy_zyte_api/addon.py
wRAR: I think the code I added (which uses priorities) is enough for this, and we have a test for that.

scrapy_zyte_api/addon.py
kmike: The approach for http(s) handlers sounds good to me - fallback to default handler, don't inherit. What are your concerns about fragility, what possible issues do you see?

wRAR: My concerns were mostly "what if Scrapy changes internal handling of download handlers, e.g. adds some explicit cleanup for ones in `Downloader.handlers`", but such cleanup already exists: `scrapy.core.downloader.handlers.DownloadHandlers` calls `.close()` for each handler it contains in the `engine_stopped` handler. So we already need to call `.close()` on our internal handler instance manually, and if that part of Scrapy changes then also change something.

wRAR: As a first step for this I changed the existing feature of the hardcoded fallback, it looks like it works, now we need to decide how to pass the user-set handler to our code, as when the handler is loaded it cannot know the old overwritten setting value. I don't like the idea of setting a special setting in the addon code but I'm not sure what other ways we have to pass this.

kmike: Can we handle it in the same way as request fingerprints? Before changing the handlers, addon reads the current value of the handler settings, and puts it to a setting. Download handler accesses the setting and loads the old, now overwritten, handler as a fallback.

wRAR: That's the solution I said I don't like :)

kmike: Ok :)

wRAR: > Can we handle it in the same way as request fingerprints? 

Also, if by this you mean `ZYTE_API_FALLBACK_REQUEST_FINGERPRINTER_CLASS` then we currently only tell the user to set it when they want, we don't set it outrselves. Should the addon set it to the old value of `REQUEST_FINGERPRINTER_CLASS`? Or maybe only if it's not set yet?

kmike: > Also, if by this you mean ZYTE_API_FALLBACK_REQUEST_FINGERPRINTER_CLASS then we currently only tell the user to set it when they want, we don't set it outrselves. Should the addon set it to the old value of REQUEST_FINGERPRINTER_CLASS? 

Yes, that's what I was thinking. 

> Or maybe only if it's not set yet?

Yes, it seems reasonable.

wRAR: We may need to modify Scrapy download handler code after all: it looks like a handler doesn't know which scheme it handles, and so it doesn't know which fallback handler to use.

scrapy_zyte_api/handler.py
```diff
@@ -143,13 +142,32 @@ def _build_client(settings):
...
     def download_request(self, request: Request, spider: Spider) -> Deferred:
         api_params = self._param_parser.parse(request)
         if api_params is not None:
             return deferred_from_coro(
                 self._download_request(api_params, request, spider)
             )
-        return self.fallback_handler.download_request(request, spider)
+        if not self._fallback_handler:
+            scheme = urlparse_cached(request).scheme
+            self._load_fallback_handler(scheme)
```
kmike, line 168: What do you think about defining 2 handler classes, for http and https? In this case they will know which scheme they handle.

it seems we can still advice users to use a single handler for both schemas if they don't use addon, because handler only needs to know schema to use the _ZYTE_API_FALLBACK_HTTP(S)_HANDLER options, which are private.

kmike: I guess the question is what is better :) What do you prefer?

wRAR: I think having two additional classes that only contain several lines of code is fine here, I'll try to implement that and see how it looks.

scrapy_zyte_api/addon.py
```diff
@@ -0,0 +1,66 @@
...
+        ):
+            return "'https' value in the 'DOWNLOAD_HANDLERS'"
+        if (
+            settings.get("REQUEST_FINGERPRINTER_CLASS")
+            != "scrapy.utils.request.RequestFingerprinter"
+        ):
+            return "'REQUEST_FINGERPRINTER_CLASS'"
+        return None
+
+    def update_settings(self, settings: BaseSettings) -> None:
```
kmike, line 28: By the way, what do you think about configuring SCRAPY_POET_PROVIDERS as well?

kmike: It's a bit tricky though, because it's not clear how to un-configure it if user doesn't want it.

README.rst
```diff
@@ -62,13 +67,24 @@ To enable this plugin:
     Scrapy setting to
     ``"twisted.internet.asyncioreactor.AsyncioSelectorReactor"``.
 
+In both cases:
+
 -   Set `your Zyte API key
     <https://docs.zyte.com/zyte-api/usage/general.html#authorization>`_ as
     either the ``ZYTE_API_KEY`` Scrapy setting or as an environment variable of
     the same name.
```
kmike, line 75: If we don't auto-configure SCRAPY_POET_PROVIDERS, I think we should mention it here (and also in the example below).

scrapy_zyte_api/handler.py
wRAR: I think it looks fine, though I needed to also make a 3rd class for the normal use case. It also no longer reads the default settings and just hardcodes the default handler, like in the current master.

README.rst
wRAR: It's mentioned, I just need to merge main

kmike: right :)

scrapy_zyte_api/addon.py
wRAR: Yeah, I tried this and as there are no priorities in the dict and the addon code runs later than the settings from `settings.py` are loaded you can't disable it via `settings.py`.

wRAR: We could add an addon-specific setting which would tell the addon to not add the provider, not sure if this is a good way to do it. In any case it's a good example of what would addon authors want.

README.rst
```diff
@@ -100,14 +105,43 @@ To enable this plugin:
...
+    DOWNLOAD_HANDLERS = {
+        "http": "scrapy_zyte_api.ScrapyZyteAPIDownloadHandler",
+        "https": "scrapy_zyte_api.ScrapyZyteAPIDownloadHandler",
+    }
+    DOWNLOADER_MIDDLEWARES = {
+        "scrapy_zyte_api.ScrapyZyteAPIDownloaderMiddleware": 1000,
+    }
+    REQUEST_FINGERPRINTER_CLASS = "scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter"
+    TWISTED_REACTOR = "twisted.internet.asyncioreactor.AsyncioSelectorReactor"
+    ZYTE_API_KEY = "YOUR_API_KEY"
```
kmike, line 137: The readme structure changed a bit, so now this example is a duplicate of the quickstart section. What do you think about moving ADDONS... example to the quickstart as well?

scrapy_zyte_api/addon.py
```diff
@@ -0,0 +1,66 @@
+from typing import Optional
+
+from scrapy.settings import BaseSettings
+
+from scrapy_zyte_api import ScrapyZyteAPIDownloaderMiddleware
+
+
+class Addon:
+    @staticmethod
+    def _check_settings(settings: BaseSettings) -> Optional[str]:
```
kmike, line 10: is this method used?

scrapy_zyte_api/addon.py
wRAR: If we provide a setting to disable the provider I wonder how often will it be used, because it one line of code in settings.py to enable the provider manually.

kmike: Yes, it doesn't seem to worth it.

README.rst
```diff
@@ -55,7 +55,15 @@ Get a `Zyte API`_ key, and add it to your project settings.py:
...
 ``ZYTE_API_KEY`` environment variable.
 
-Then, set up the scrapy-zyte-api integration:
+Then, set up the scrapy-zyte-api integration. If your Scrapy version supports addons:
+
+.. code-block:: python
+
+    ADDONS = {
+        "scrapy_zyte_api.Addon": 1,
+    }
```
kmike, line 64: There is a note about transparent mode later; I think we should make sure it's written in a way which makes it clear it doesn't apply for addon config.

wRAR: Done, though I'm not sure about the wording.

README.rst
```diff
@@ -921,3 +948,30 @@ Running behind a proxy
...
+When the plugin needs to do requests directly, not via Zyte API, it uses a
+fallback download handler. It will use the Scrapy default one unless the
+``ZYTE_API_FALLBACK_HTTP_HANDLER`` and ``ZYTE_API_FALLBACK_HTTPS_HANDLER``
+settings are set. So if you want to use some other handler as a fallback, set
+those variables::
+
+    ZYTE_API_FALLBACK_HTTP_HANDLER = "myproject.MyHandler"
+    ZYTE_API_FALLBACK_HTTPS_HANDLER = "myproject.MyHandler"
+
+Alternatively, if you use the Scrapy addon, you can set the
```
kmike, line 965:
```suggestion
Alternatively, if you use the Scrapy addon, you can keep the
```

kmike: If users are using cusotm downloade handlers, they already should be in their settings.

Gallaecio (approved):

wRAR (commented):
Looks good to me

PR #100, opened 2023-07-12
Support more data types in the provider

kmike:
The PR adds support for other data types in the provider.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -863,6 +863,11 @@ The currently supported dependencies are:
 * ``web_poet.BrowserHtml``
 * ``web_poet.BrowserResponse``
 * ``zyte_common_items.Product``
+* ``zyte_common_items.ProductList``
+* ``zyte_common_items.ProductNavigation``
+* ``zyte_common_items.Article``
+* ``zyte_common_items.ArticleList``
+* ``zyte_common_items.ArticleNavigation``
 
 The provider will make a request to Zyte API using the ``ZYTE_API_KEY`` and
 ``ZYTE_API_URL`` settings. It will ignore the transparent mode and parameter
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -19,7 +19,7 @@
         "provider": [
             "scrapy-poet>=0.9.0",
             "web-poet>=0.13.0",
-            "zyte-common-items",
+            "zyte-common-items>=0.7.0",
```
kmike, line 22: this is a bit tricky, because scrapy-zyte-api supports Python 3.7, but recent zyte-common-items don't. I think that's fine, because this dependency is optional.

scrapy_zyte_api/providers.py
```diff
@@ -47,7 +63,14 @@ async def __call__(
...
 
         html_requested = BrowserResponse in to_provide or BrowserHtml in to_provide
-        item_keywords = {Product: "product"}
+        item_keywords = {
+            Product: "product",
+            ProductList: "productList",
+            ProductNavigation: "productNavigation",
+            Article: "article",
+            ArticleList: "articleList",
+            ArticleNavigation: "articleNavigation",
```
kmike, line 72: I haven't added tests for this, because it doesn't seem they add much, as compared to the existing Product test.

README.rst
```diff
@@ -30,6 +30,10 @@ Requirements
 * Python 3.7+
 * Scrapy 2.0.1+
 
+Scrapy-poet integration requires more recent software:
```
Gallaecio, line 33:
```suggestion
scrapy-poet integration requires more recent software:
```

Gallaecio (approved):

setup.py
wRAR: Should we just drop Python 3.7 support?

kmike: Let's not be aggressive with deprecating old software here, to aid the migration from SPM to Zyte API.

PR #101, opened 2023-07-12
Cleanup tools configuration

kmike:
* move to pyproject.toml when possible
* use pre-commit in tox, don't duplicate configuration
* don't run mypy in pre-commit
* update flake8, mypy and isort
* update github actions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ profile = "black"
 multi_line_output = 3
 
 [tool.mypy]
-#check_untyped_defs = true
+check_untyped_defs = true
 ignore_missing_imports = true
 no_warn_no_return = true
 files = [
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[tool.black]
+target-version = ["py38", "py39", "py310", "py311"]
+
+[tool.isort]
+profile = "black"
+multi_line_output = 3
+
+[tool.mypy]
+#check_untyped_defs = true
```
Gallaecio, line 9: To be addressed later on, or a leftover?

Gallaecio (approved):

pyproject.toml
kmike: Oh, it's used actually :) Let me fix it

PR #110, opened 2023-07-14
[wip] Actions dependency

kmike:
It's also an option to add this dependency to web-poet, as ZyteApiActions.

The implementation is a bit tricky; this is the first time we have a parametrized dependency. It might require changes in andi (https://github.com/scrapinghub/andi/pull/24) and in other packages (scrapy-poet?). Only Python 3.9+ is supported here.

### DIFF
--- a/scrapy_zyte_api/page_inputs.py
+++ b/scrapy_zyte_api/page_inputs.py
@@ -52,4 +52,4 @@ def Actions(actions: List[_Action]):
             # ...
 
     """
-    return Annotated[_Actions, _ActionsList[actions]]
+    return Annotated[_Actions, _ActionsList(actions)]
--- a/scrapy_zyte_api/page_inputs.py
+++ b/scrapy_zyte_api/page_inputs.py
@@ -39,11 +39,17 @@ def Actions(actions: List[_Action]):
     """
     Add this dependency to your page object to execute Zyte API actions::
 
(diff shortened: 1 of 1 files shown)


scrapy_zyte_api/page_inputs.py
```diff
@@ -0,0 +1,55 @@
...
+    urlMatchingOptions: Optional[str]
+    selector: Optional[_Selector]
+    onError: Optional[str]
+    delay: Optional[float]
+    timeout: Optional[float]
+    id: Optional[str]
+    keyword: Optional[str]
+    key: Optional[str]
+    source: Optional[str]
+    options: Optional[dict]
```
Gallaecio, line 25: Wouldn’t it be better to have an approach that does not require defining action top-level fields here?

kmike: This section is not required; the code would work fine without it as-is. If there are other top-level fields, I think it also should work as-is. I added it to have some type annotations for the users, via TypedDict. Would you prefer to have it not type checked?

Gallaecio: I was not familiar with `TypedDict`, nice.

wRAR:
Superseded by https://github.com/scrapy-plugins/scrapy-zyte-api/pull/181

Issue #111, opened 2023-07-14
A way to limit the Zyte API requests


BurnzZ:
I'm thinking of a feature where the user can indicate the max number of Zyte API requests for a given spider crawl. When this number is reached, the spider shuts down.

The main use case for this is to limit the costs per crawl.

Gallaecio:
Wouldn’t [CLOSESPIDER_PAGECOUNT](https://doc.scrapy.org/en/latest/topics/extensions.html#closespider-pagecount) do the job in most cases?

BurnzZ:
Yes, that's right. But it would also include non-ZAPI requests to the count.

BurnzZ:
An alternative: https://github.com/croqaz/scrapy-count-filter

BurnzZ:
Closing as https://github.com/scrapy-plugins/scrapy-zyte-api/pull/114 has been merged

PR #114, opened 2023-07-19
Introduce ZYTE_API_MAX_REQUESTS

BurnzZ:
An attempt on https://github.com/scrapy-plugins/scrapy-zyte-api/issues/111 to gather feedback.

This has minimal changes that only addresses the limits on ZAPI Requests.

In the future, we also might want to impose other limits like the following which may affect the implementation:
- item count
- per data type 
- per page type
- etc

TODO
- [x] README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -723,6 +723,19 @@ as HTTP 520 errors, you can implement:
 .. _tenacity.AsyncRetrying: https://tenacity.readthedocs.io/en/latest/api.html#tenacity.AsyncRetrying
 
 
+Misc Settings
+=============
+
+- ``ZYTE_API_MAX_REQUESTS``
+
+  default: None
+
+  When set to an integer value > 0, the spider will close when the number of
+  successful Zyte API requests reaches it. Note that in some cases, the actual
+  number of successful Zyte API requests would be below this number if some of
+  the in-progress requests fail or error out.
+
+
 Stats
 =====
 
--- a/README.rst
+++ b/README.rst
@@ -723,12 +723,12 @@ as HTTP 520 errors, you can implement:
 .. _tenacity.AsyncRetrying: https://tenacity.readthedocs.io/en/latest/api.html#tenacity.AsyncRetrying
 
 
-Misc Settings
+Misc settings
 =============
 
 - ``ZYTE_API_MAX_REQUESTS``
 
-  default: None
+  Default: ``None``
 
   When set to an integer value > 0, the spider will close when the number of
   successful Zyte API requests reaches it. Note that in some cases, the actual
--- a/scrapy_zyte_api/_downloader_middleware.py
+++ b/scrapy_zyte_api/_downloader_middleware.py
@@ -1,4 +1,10 @@
+import logging
+
+from scrapy.exceptions import IgnoreRequest
+
 from ._params import _ParamParser
+
+logger = logging.getLogger(__name__)
 
 
 class ScrapyZyteAPIDownloaderMiddleware:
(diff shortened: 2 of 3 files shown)


scrapy_zyte_api/_downloader_middleware.py
```diff
@@ -23,3 +36,22 @@ def process_request(self, request, spider):
...
+            raise IgnoreRequest("Reached max Zyte API requests")
+
+    def _max_requests_reached(self, downloader) -> bool:
+        if not self._max_requests:
+            return False
+
+        zapi_req_count = self._crawler.stats.get_value("scrapy-zyte-api/processed", 0)
+        download_req_count = sum(
+            [
+                len(slot.transferring)
```
BurnzZ, line 51: This is a tricky one since if one of the ongoing requests fail (after some best efforts to have them retried), then the _successful billable ZAPI requests_ **<** `ZYTE_API_MAX_REQUESTS`.

Perhaps this would be okay?

wRAR: Do you mean "the user wanted 1000 requests but only used 990"? I think that's fine.

scrapy_zyte_api/_downloader_middleware.py
```diff
@@ -23,3 +36,22 @@ def process_request(self, request, spider):
             request.meta["download_slot"] = slot_id
         _, slot = downloader._get_slot(request, spider)
         slot.delay = 0
+
+        if self._max_requests_reached(downloader):
+            self._crawler.engine.close_spider(spider, "closespider_max_zapi_requests")
+            raise IgnoreRequest("Reached max Zyte API requests")
```
wRAR, line 42: As, I think, this code will be executed several times, I wonder how will it look in the log, hopefully not too ugly.

BurnzZ: There should just be a few of them since the spider is immediately closed.

scrapy_zyte_api/_downloader_middleware.py
BurnzZ: where they wanted exactly 1000 billable ZAPI requests but only 990 were billed as 10 had errors in them.

wRAR:
I think the approach is fine, though I haven't thought about the `download_req_count` calculation.

scrapy_zyte_api/_downloader_middleware.py
Gallaecio: As long as the message reads “this request is skipped because the max has been reached” rather than just “the max has been reached”, I think it is OK for this to appear several times.

BurnzZ: Message updated in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/114/commits/660cdb2ab5cf04c30c058c6f471f1ad8044e5876

BurnzZ:
Hey folks, I've unmarked this one as _draft_.

I think we can think about other things to set as max like item count, per data type, per page type, etc in another future PR.

README.rst
```diff
@@ -723,6 +723,19 @@ as HTTP 520 errors, you can implement:
 .. _tenacity.AsyncRetrying: https://tenacity.readthedocs.io/en/latest/api.html#tenacity.AsyncRetrying
 
 
+Misc Settings
+=============
```
Gallaecio, line 727:
```suggestion
Misc settings
=============
```

README.rst
```diff
@@ -723,6 +723,19 @@ as HTTP 520 errors, you can implement:
...
+
+- ``ZYTE_API_MAX_REQUESTS``
+
+  default: None
```
Gallaecio, line 731:
```suggestion
  Default: ``None``
```

Gallaecio (approved):

scrapy_zyte_api/_downloader_middleware.py
kmike: This behavior sounds fine to me.

kmike:
Thanks @BurnzZ!

PR #117, opened 2023-08-04
CHANGES.rst: Cover 0.11.0

Gallaecio:


CHANGES.rst
```diff
@@ -1,6 +1,15 @@
 Changes
 =======
 
+0.11.0 (to be released)
+-----------------------
+
+* Added a ``ZYTE_API_MAX_REQUESTS`` setting to limit the number of Zyte API
+  requests that a spider can send. Reaching the limit stops the spider.
```
kmike, line 8:
```suggestion
  successful requests that a spider can send. Reaching the limit stops the spider.
```

BurnzZ: `ZYTE_API_MAX_PROCESSED_REQUESTS` ?

If the max requests was 10, and the spider already closed the engine at 8 successful requests and 2 are still waiting for a response, if these 2 remaining requests fail (after retries) then there'd only be 8 successful requests which are billed. Perhaps adding "processed" could help remove the confusion?

kmike: ZYTE_API_MAX_RESPONSES?

Gallaecio: I slightly prefer the existing `ZYTE_API_MAX_REQUESTS` over those 2 alternatives, but no strong opinion.

As long as we do not plan on adding a future setting that can be confused with this one, and we explain accurately how it works in the docs as we do, I think it is OK for the setting name not to cover all the semantic nuances of the setting.

BurnzZ: `ZYTE_API_MAX_RESPONSES` could also bear the idea where 429 could be included in it.

`ZYTE_API_MAX_SUCCESSFUL_RESPONSES`? `ZYTE_API_MAX_BILLED_RESPONSES`?

kmike (approved):

CHANGES.rst
kmike: I'm also fine with keeping it as-is.

PR #122, opened 2023-08-08
Test the minimum dependencies of the provider extra

Gallaecio:
Note: I was hoping this would have caused the accidental scrapy-poet dependency upgrade to fail, but it turns out the test would simply have upgraded the dependency while installing the package inside the tox environment, and passed nonetheless.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -80,6 +80,17 @@ deps =
 [testenv:provider]
 extras = provider
 
+[testenv:pinned-provider]
+# zyte-common-items ≥ 0.6.0 requires Python ≥ 3.8.
+basepython=python3.8
+extras = provider
+deps =
+    # scrapy-poet 0.10.0 depends on scrapy>=2.6.0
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -80,6 +80,17 @@ deps =
...
+[testenv:pinned-provider]
+# zyte-common-items ≥ 0.6.0 requires Python ≥ 3.8.
+basepython=python3.8
+extras = provider
+deps =
+    # scrapy-poet 0.10.0 depends on scrapy>=2.6.0
+    {[testenv:pinned-scrapy-2x6]deps}
+    scrapy-poet==0.10.0
+    web-poet==0.13.0
+    zyte-common-items==0.7.0
```
kmike, line 92: Could you please add a comment to setup.py, to ensure we don't forget to update tox.ini when the minimum dependencies change?

kmike (approved):

PR #128, opened 2023-08-25
Track top-level request args in stats

Gallaecio:
I thought about excluding `url` and `jobId`, but maybe that could be counterproductive, if some user may think they are failing to be sent because of the stats.

Similarly, I wonder if unwrapping some specific arguments makes more sense. For example, do not track `experimental`, but its nested fields (e.g. `experimental.requestCookies`).

I also wonder if I should be implementing this in python-zyte-api first. And, if so, if we want to display them in the command-line output somehow.


kmike:
I think it'd be nice to have special treatment for the experimental features. Other than that, the approach looks fine to me.

wRAR (approved):

kmike:
Nice!

PR #130, opened 2023-09-21
Identify scrapy-zyte-api usage via custom user-agent

PyExplorer:
Goals:
- It should also allow to change the user agent, e.g. for zyte-crawlers use case (not implemented yet).
- In the user agent we should keep the information about python-zyte-api version used (possible with the fix).

This is one of the options to set custom user-agent for scrapy-zyte-api and keep  for python-zyte-api. But it requires changes in client.py python-zyte-api https://github.com/zytedata/python-zyte-api/blob/main/zyte_api/aio/client.py#L60 like:
```
    async def request_raw(self, query: dict, *,
                          endpoint: str = 'extract',
                          session=None,
                          handle_retries=True,
                          retrying: Optional[AsyncRetrying] = None,
                          ):
        retrying = retrying or self.retrying
        post = _post_func(session)
        auth = aiohttp.BasicAuth(self.api_key)
        headers = {
            'User-Agent': user_agent(aiohttp, extra_user_agent=query.pop("user-agent", None)),
            'Accept-Encoding': 'br'
        }
        response_stats = []
        start_global = time.perf_counter()

```
Some other options to provide the way to send user-agent from zyte-crawlers:
1.Use settings with name of package and set them during creating client (requires changes in AsyncClient)
2. Use cb_kwargs or meta - but looks like they are cleaned during the request process

It's good to discuss all above.


Gallaecio (commented):
It is probably best to pass the user agent to `AsyncClient` as a parameter instead.

PyExplorer:
> It is probably best to pass the user agent to `AsyncClient` as a parameter instead, once python-zyte-api supports that.

@Gallaecio is it ok to change AsyncClient this way? 
```
    def __init__(self, *,
                 api_key=None,
                 api_url=API_URL,
                 n_conn=15,
                 retrying: Optional[AsyncRetrying] = None,
                 custom_user_agent=None
                 ):
```

Gallaecio:
Sorry, accidentally edited your comment instead of answering :facepalm: 

Answer: I think so, yes. Maybe even call it just `user_agent`, if the value is not None then it is custom :slightly_smiling_face: .

PyExplorer:
> It is probably best to pass the user agent to `AsyncClient` as a parameter instead, once python-zyte-api supports that.

aha, ok, with this. @Gallaecio How do you think we can send user-agent from zyte-crawlers? I was thinking about adding user-agent for python-zyte-api here https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/providers.py#L81 with meta (as only spider with zyte-crawlers go here), but probably it's better to send this right from zyte-crawlers repo with settings or else - any ideas?

PyExplorer:
@Gallaecio could you take a look?
Related PR in `python-zyte-api`
https://github.com/zytedata/python-zyte-api/pull/50

PyExplorer:
@Gallaecio @kmike could you take a look?

scrapy_zyte_api/handler.py
```diff
@@ -129,6 +131,7 @@ def _build_client(settings):
                 api_key=settings.get("ZYTE_API_KEY") or None,
                 api_url=settings.get("ZYTE_API_URL") or API_URL,
                 n_conn=settings.getint("CONCURRENT_REQUESTS"),
+                user_agent=f'{settings.get("_USER_AGENT", "")} {USER_AGENT} {PYTHON_ZYTE_API_USER_AGENT}'.lstrip(),
```
kmike, line 134: Let's name the setting _ZYTE_API_USER_AGENT

kmike: It also looks like _ZYTE_API_USER_AGENT should overwrite the user_agent completely, like in python-zyte-api.

kmike: Alternatively, you can call the setting _ZYTE_API_USER_AGENT_EXTRA or something like that, though overwriting completely sounds a bit better to me

scrapy_zyte_api/utils.py
```diff
@@ -1,6 +1,10 @@
+from importlib.metadata import version
```
kmike, line 1: TIL

kmike: I'm not sure we can use it though, https://docs.python.org/3/library/importlib.metadata.html says it's 3.8+, while scrapy-zyte-api declares Python 3.7 support.

scrapy_zyte_api/handler.py
Gallaecio: What about calling it `ZYTE_API_USER_AGENT`, without the `_` prefix, and simply not document it?

scrapy_zyte_api/utils.py
Gallaecio: > scrapy-zyte-api declares Python 3.7 support

That has an easy fix :)

scrapy_zyte_api/handler.py
kmike: My concern is that users might discover it, and think it's a user-agent to send to the remote websites, similar to USER_AGENT setting.

kmike: Underscore makes it look a bit more scrary at least :)

Gallaecio: Love “scrary” as a concept :slightly_smiling_face:

PyExplorer:
@kmike @Gallaecio @wRAR @BurnzZ could you take a look at the PR with last changes from the discussion about version?

scrapy_zyte_api/__version__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.11.1"
```
Gallaecio, line 1: :sweat_smile: 

```suggestion
__version__ = "0.12.0"
```

Gallaecio:
I think it was OK to leave setup.py the way it was and have bump2version configuration edit an extra file instead of a different file, but I am OK with the new approach as well.

The only thing left is Python 3.7 support, I think.

wRAR:
`pinned-provider` also failed, with "cannot import name 'USER_AGENT' from 'zyte_api.utils'"

scrapy_zyte_api/__version__.py
PyExplorer: Oh, haven't noticed that it was updated, thanks @Gallaecio !

PyExplorer:
> I think it was OK to leave setup.py the way it was and have bump2version configuration edit an extra file instead of a different file, but I am OK with the new approach as well.
> 
> The only thing left is Python 3.7 support, I think.

Oh, actually, I've made it as in python-zyte-api just for consistency

PyExplorer:
> `pinned-provider` also failed, with "cannot import name 'USER_AGENT' from 'zyte_api.utils'"

@wRAR what do you think will be the best option to fix it:
- set "zyte-api>=0.4.7" in setup.py
- fix test by importing and using only __version__ from ``python-zyte-api like``
```
from zyte_api.utils import __version__ as zyte_api_version
...
   "user_agent,expected",
    (
        (
            None,
            f"{USER_AGENT} python-zyte-api/{zyte_api_version}",
        ),
        (
            "zyte-crawlers/0.0.1",
            "zyte-crawlers/0.0.1",
        ),
    ),

```

scrapy_zyte_api/utils.py
```diff
@@ -1,6 +1,10 @@
 import scrapy
 from packaging.version import Version
 
+from .__version__ import __version__
+
+USER_AGENT = f"scrapy-zyte-api/{__version__}"
```
kmike, line 6: I think it might be a bit confusing, because that't not actually the user agent which is used, as the real user agent includes python-zyte-api version.

scrapy_zyte_api/handler.py
```diff
@@ -129,6 +131,8 @@ def _build_client(settings):
                 api_key=settings.get("ZYTE_API_KEY") or None,
                 api_url=settings.get("ZYTE_API_URL") or API_URL,
                 n_conn=settings.getint("CONCURRENT_REQUESTS"),
+                user_agent=settings.get("_ZYTE_API_USER_AGENT")
+                or f"{USER_AGENT} {PYTHON_ZYTE_API_USER_AGENT}",
```
kmike, line 135: nitpick: would it make sense to use `settings.get("_ZYTE_API_USER_AGENT", default=...)` instead?

PyExplorer: @kmike I see that default for settings.get is None - do you think it worth to set this explicitly here?

scrapy_zyte_api/utils.py
PyExplorer: @kmike what about rename this to SCRAPY_ZYTE_API_USER_AGENT? A little long though...

kmike: It seems you can store the complete user agent as a constant, why do you need a constant with a part of the user agent?

scrapy_zyte_api/handler.py
kmike: I mean that instead of `settings.get("_ZYTE_API_USER_AGENT") or ...` you can use `settings.get("_ZYTE_API_USER_AGENT", default=...)` - or is there any issue with that?

PyExplorer: @kmike ah, got it, right, thanks!

scrapy_zyte_api/utils.py
PyExplorer: you are right @kmike , that's better, thanks - fixed that. One question left here https://github.com/scrapy-plugins/scrapy-zyte-api/pull/130#issuecomment-1737095570

scrapy_zyte_api/handler.py
Gallaecio: The `or` approach usually has the advantage that it lets you use an empty string on the command line to unset the value (e.g. if set at project level), but for this specific setting I guess that’s not much of a concern.

scrapy_zyte_api/utils.py
kmike: > One question left here https://github.com/scrapy-plugins/scrapy-zyte-api/pull/130#issuecomment-1737095570

I think that's better to bump python-zyte-api min version requirement. We want users to use the latest client library anyways, because it may contain fixes for the client library behavior.

Gallaecio (approved):

wRAR (approved):

.bumpversion.cfg
```diff
@@ -4,6 +4,4 @@ commit = True
 tag = True
 tag_name = {new_version}
 
-[bumpversion:file:setup.py]
-search = version="{current_version}",
-replace = version="{new_version}",
+[bumpversion:file:scrapy_zyte_api/__version__.py]
```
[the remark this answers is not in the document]
PyExplorer: I've checked this locally - after running bump2version - two files were changed with the new version:
__version__.py (version was changed)
.bumpversion.cfg (current_version was changed)
Looks good.
Thanks for the suggestion! @wRAR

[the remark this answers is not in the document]
PyExplorer: @kmike @BurnzZ could you take a look at the more or less final version?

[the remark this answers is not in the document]
Gallaecio: For the record, the reason the prior config had 2 additional settings was because it involved `setup.py`, where there was a chance other version numbers (those of dependencies) would accidentally get upgraded if they ever matched the package version. That is avoided by forcing the whole `version="…"` pattern to be searched for in `setup.py`, instead of just the version number. That’s not needed for `__version__.py`, though.

BurnzZ (approved):

kmike (approved):

PR #131, opened 2023-09-21
Fix typing issues for typed Scrapy.

wRAR:


scrapy_zyte_api/responses.py
```diff
@@ -188,8 +192,8 @@ def _process_response(
 
     if api_response.get("httpResponseHeaders") and api_response.get("httpResponseBody"):
         response_cls = responsetypes.from_args(
-            headers=api_response["httpResponseHeaders"],
-            url=api_response["url"],
+            headers=cast(List[Dict[str, str]], api_response["httpResponseHeaders"]),
```
wRAR, line 195: This produces an error which looks valid: `api_response["httpResponseHeaders"]` is a list of dicts with "name" and "value" keys while `ResponseTypes.from_headers` expects a `Mapping[bytes, bytes]` (not even strings).

Gallaecio (approved):

scrapy_zyte_api/responses.py
Gallaecio: We’ll need to address that.

wRAR:
Scrapy < 2.1 has a hidden bug, accidentally fixed in https://github.com/scrapy/scrapy/pull/4481: `scrapy.responsetypes.ResponseTypes.from_headers()` expects `headers` keys to be case-insensitive so it can't work with simple dicts. Looks like we need to skip a couple of tests on Scrapy 2.0.1.

wRAR:
OTOH I think this breakage will also happen in user code. Unless we want to drop Scrapy 2.0.1, we need to use `scrapy.http.Headers` here.

scrapy_zyte_api/addon.py
```diff
@@ -74,7 +76,7 @@ def update_settings(self, settings: BaseSettings) -> None:
             settings.set(
                 "REQUEST_FINGERPRINTER_CLASS",
                 "scrapy_zyte_api.ScrapyZyteAPIRequestFingerprinter",
-                settings.getpriority("REQUEST_FINGERPRINTER_CLASS"),
+                cast(int, settings.getpriority("REQUEST_FINGERPRINTER_CLASS")),
```
kmike, line 79: is it a Scrapy typing issue that getpriority doesn't annotated as returning int?

wRAR: It can be None. IIRC I decided that a cast is more simple than extracting the var and checking it.

kmike: Is it an error if the priority is None? Or is it guaranteed we'll get an int? Sorry, I'm a bit confused about the priority :) If it's None, wouldn't settings.set allow it? If no, how would you get None in a first place?

wRAR: `getpriority()` returns None for unset settings.

tests/test_sessions.py
```diff
@@ -393,7 +393,7 @@ class UseChecker(ConstantChecker):
     def check(self, response: Response, request: Request) -> bool:
         if response.meta.get(SESSION_INIT_META_KEY, False) is True:
             return True
-        return super().check(request, response)
+        return super().check(response, request)
```
Gallaecio, line 396: :open_mouth:

wRAR: Yeah.

It's a test though.

PR #132, opened 2023-09-26
Drop the Set-Cookie header

Gallaecio:


wRAR (approved):

scrapy_zyte_api/responses.py
```diff
@@ -22,7 +22,11 @@ class ZyteAPIMixin:
         # Zyte API already decompresses the HTTP Response Body. Scrapy's
         # HttpCompressionMiddleware will error out when it attempts to
         # decompress an already decompressed body based on this header.
-        "content-encoding"
+        "content-encoding",
+        # Cookies should be fetched from experimental.responseCookies only, the
+        # Set-Cookie header may define cookies from the main HTTP response that
+        # were overriden later on, e.g. during browser rendering.
+        "set-cookie",
```
proway2, line 29: I don't think we should remove headers when `httpResponseBody = True`.

wRAR (approved):

scrapy_zyte_api/responses.py
```diff
@@ -90,18 +91,30 @@ def _prepare_headers(cls, api_response: Dict[str, Any]):
         input_headers: Optional[List[Dict[str, str]]] = api_response.get(
             "httpResponseHeaders"
         )
+        response_cookies: Optional[List[Dict[str, str]]] = api_response.get(
+            "experimental", {}
+        ).get("responseCookies")
         if input_headers:
+            headers_to_remove = copy(cls.REMOVE_HEADERS)
+            if response_cookies or "httpResponseBody" not in api_response:
```
kmike, line 99: Does it handle an edge case where only httpResponseHeaders are requested, without httpResponseBody?

Gallaecio: No it does not, did not even cross my mind.

Gallaecio: Hmm… This is more involved than I initially thought.

Since `httpResponseHeaders` can be combined with any other output, including extraction (current and future extraction keys), filtering only by “`httpResponseHeaders` and no other output” is not feasible.

So I think we need to choose:
- Keep the logic as is, remove `Set-Cookie` even if `httpResponseHeaders` is the only output, as long as response cookies are requested as well.
- Keep the `Set-Cookie` header whenever `httpResponseHeaders` is present, even if `browserHtml` or other outputs are also requested.
- Keep the `Set-Cookie` header whenever `httpResponseHeaders` is present as long as no other _known_ output is requested, and accept that there will be times between Zyte API adding support for a new extraction type and us supporting it here where `Set-Cookie` would be deleted for those new scenarios. We could expose the list of extractions as a setting as a workaround, if we want to go that far.

kmike: Hm, so, we need to remove the header if we know that the browser rendering was not used internally. And it's not straightforward to know if it was used or not: e.g. it could be the case httpResponseBody + browserHtml could be supported at the same time.

Is it the case that ideally, we'd like to keep the Set-Cookie header always (for informational purposes), but ensure it's not processed by the cookie middleware when responseCookies are present? Or are there any issues with that? Can we address this directly somehow, instead of removing Set-Cookie header conditionally?

Also, what's the actual use case of keeping Set-Cookie header when responseCookies are received? If there is no use case, and addressing the point above is complicated, then can we drop Set-Cookie unconditionally when responseCookies are present?

Gallaecio: That still does not explain why, when using `httpResponseBody`, you would read the cookies from `Set-Cookie` instead of `responseCookies`. It’s accommodating that scenario, i.e. not dropping `Set-Cookie` even though `responseCookies` have been received, what makes the implementation tricky.

proway2: This is what stated on our website for `httpResponseHeaders`:

> The Set-Cookie header value, when present, contains the header value received from the main HTTP response. These cookies could have changed later on, e.g. during browser rendering. Usually you will want to ignore this header in favor of [responseCookies](https://docs.zyte.com/zyte-api/usage/reference.html#operation/extract/response/200/responseCookies), which provides the final cookies.

I'm not sure what `Usually` exactly means here, I'd read cookies from `responseCookies`. But I guess `responseCookies` may be set to `False` but `httpResponseHeaders` may be set to `True`, it may be the case when a user needs something from headers (e.g. some key) but doesn't actually need a cookie.

I'd say that if `browserHtml` is used, remove `httpResponseHeaders` at all. If `httpResponseBody` is used and cookies are enabled and `responseCookies` is `True` - read cookies from `responseCookies`, but if `responseCookies` is `False` - read cookies from `httpResponseHeaders`. In any case any other header from `httpResponseHeaders` must remain untouched.

kmike:
> The problem is that when httpResponseHeaders is used together with browserHtml it's obviously a mistake.
> I'd say that if browserHtml is used, remove httpResponseHeaders at all.

I'm not sure about this; if it's supported by the API, and requested explicitly by the user, then it might not be a mistake.

Overall, it seems we're all on the same page - we should prioritize responseCookies, and there is no issue with removing Set-Cookie from httpResponseHeaders if responseCookies are present, right?

scrapy_zyte_api/responses.py
kmike: > The problem is that when httpResponseHeaders is used together with browserHtml it's obviously a mistake.
> I'd say that if browserHtml is used, remove httpResponseHeaders at all.

I'm not sure about this; if it's supported by the API, and requested explicitly by the user, then it might not be a mistake.

Overall, it seems we're all on the same page - we should prioritize responseCookies, and there is no issue with removing Set-Cookie from httpResponseHeaders if responseCookies are present, right?

proway2: > I'm not sure about this; if it's supported by the API, and requested explicitly by the user, then it might not be a mistake.

Based on the discussion in Slack though it's allowed to use httpResponseHeaders+browserHtml it's strongly discouraged. I'm not sure if there's a use case for this.

>Overall, it seems we're all on the same page - we should prioritize responseCookies, and there is no issue with removing Set-Cookie from httpResponseHeaders if responseCookies are present, right?

Yes

PR #137, opened 2023-09-28
updates for 0.12.1 release

PyExplorer:


PyExplorer:
@Gallaecio @kmike is it ok to mention ``_ZYTE_API_USER_AGENT`` in changes? And the same about adding this to readme here https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/README.rst#misc-settings?

CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Changes
 =======
 
+0.12.1 (to be released)
+-----------------------
+
+* With a new setting ``_ZYTE_API_USER_AGENT`` spider can send the custom user agent to Zyte API.
+  Otherwise, to Zyte API will be sent the custom user agent "scrapy-zyte-api/version python-zyte-api/version".
```
Gallaecio, line 8: I think we need to cover it, it being the only change in the release. Maybe something like this?

```suggestion
* The new ``_ZYTE_API_USER_AGENT`` setting allows customizing the user agent 
  string reported to Zyte API.
  
  Note that this setting is only meant for libraries and frameworks built on 
  top of scrapy-zyte-api, to report themselves to Zyte API, for client software 
  tracking and monitoring purposes. The value of this setting is *not* the 
  ``User-Agent`` header sent to upstream websites when using Zyte API.
```

PyExplorer: Looks great, thanks @Gallaecio!

PyExplorer:
@Gallaecio is it ok to be merged?

Gallaecio (approved):

PR #138, opened 2023-10-04
Remove unused extration options

kmike:
I'm not 100% sure it's a right thing to do, but it doesn't seem it hurts.
This change allows to set

```
        "ZYTE_API_PROVIDER_PARAMS": {
            "productOptions": {"extractFrom": "httpResponseBody"},
            "productNavigationOptions": {"extractFrom": "httpResponseBody"},
        }

```

option, and not having the client to pass the unused ...Options parameters to the API (which fails API validation).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -950,6 +950,18 @@ in ``settings.py``::
 
     ZYTE_API_PROVIDER_PARAMS = {"geolocation": "IE"}
 
+When ``ZYTE_API_PROVIDER_PARAMS`` setting includes one of the Zyte API
+extraction options (e.g. ``productOptions`` for ``product``), but the
+final Zyte API request doesn't include the corresponding data type, the
+unused options are automatically removed. So, it's safe to use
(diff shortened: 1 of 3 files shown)


Gallaecio (approved):
Interesting approach. Should we document it?

wRAR (approved):

wRAR:
We probably should. Not 100% sure though.

kmike:
Added some docs.

README.rst
```diff
@@ -950,6 +950,18 @@ in ``settings.py``::
 
     ZYTE_API_PROVIDER_PARAMS = {"geolocation": "IE"}
 
+When ``ZYTE_API_PROVIDER_PARAMS`` setting includes one of the Zyte API
```
Gallaecio, line 953:
```suggestion
When the ``ZYTE_API_PROVIDER_PARAMS`` setting includes one of the Zyte API
```

Gallaecio (approved):

PR #141, opened 2023-10-11
Initial support for typing.Annotated, support for extractFrom

wRAR:
Requires https://github.com/scrapinghub/andi/pull/25 and https://github.com/scrapinghub/scrapy-poet/pull/169


scrapy_zyte_api/providers.py
```diff
@@ -106,9 +127,15 @@ async def __call__(
...
-                self.update_cache(request, {item_type: item})
+
+        for cls in to_provide:
+            cls_stripped = strip_annotated(cls)
+            kw = item_keywords.get(cls_stripped)
+            if not kw:
+                continue
+            item = cls_stripped.from_dict(api_response.raw_api_response[kw])
+            if is_typing_annotated(cls):
+                item.__metadata__ = cls.__metadata__
```
wRAR, line 138: This doesn't work because `Product` has `__slots__`.

scrapy_zyte_api/providers.py
```diff
@@ -106,9 +127,15 @@ async def __call__(
...
+            results.append(item)
+            self.update_cache(request, {cls_stripped: item})
```
wRAR, line 140: This should probably use `cls`

scrapy_zyte_api/providers.py
```diff
@@ -75,9 +82,25 @@ async def __call__(
...
+                continue
+            zyte_api_meta[kw] = True
+            if not is_typing_annotated(cls):
+                continue
+            metadata = cls.__metadata__
+            if cls_stripped == Product:
+                for option in ExtractFrom:
+                    if option in metadata:
+                        product_options = zyte_api_meta.setdefault("productOptions", {})
+                        if "extractFrom" in product_options:
```
wRAR, line 99: This doesn't check if the previous value is the same, maybe we should do that instead.

tests/test_providers.py
```diff
@@ -179,3 +179,61 @@ async def test_provider_params(mockserver):
...
+
+
+@ensureDeferred
+async def test_provider_extractfrom(mockserver):
+    from typing import Annotated
+
+    @attrs.define
+    class AnnotatedProductPage(BasePage):
+        product: Annotated[Product, ExtractFrom.httpResponseBody]
+        product2: Annotated[Product, ExtractFrom.httpResponseBody]
```
wRAR, line 191: This doesn't raise a double-extractFrom exception because there is only one unique dep.

tests/test_providers.py
```diff
@@ -179,3 +179,61 @@ async def test_provider_params(mockserver):
...
+                "product": page.product,
+            }
+
+    settings = create_scrapy_settings(None)
+    settings.update(SETTINGS)
+    settings["ZYTE_API_URL"] = mockserver.urljoin("/")
+    settings["SCRAPY_POET_PROVIDERS"] = {ZyteApiProvider: 0}
+
+    item, _, _ = await crawl_single_item(AnnotatedZyteAPISpider, HtmlResource, settings)
+    assert item is None
```
wRAR, line 239: I don't see a way to catch the exception from the provider here.

scrapy_zyte_api/providers.py
wRAR: Alternatively, we should treat values from different sources with different priorities and use the top one instead of raising this error.

tests/mockserver.py
```diff
@@ -117,6 +117,14 @@ def render_POST(self, request):
                 "price": "10",
                 "currency": "USD",
             }
+            extract_from = request_data.get("productOptions", {}).get("extractFrom")
+            if extract_from:
+                from scrapy_zyte_api.providers import ExtractFrom
```
wRAR, line 122: Alternatively it can be moved to a different module (importing `providers` needs additional deps).

tests/mockserver.py
```diff
@@ -117,6 +117,14 @@ def render_POST(self, request):
...
+
+                if extract_from == ExtractFrom.httpResponseBody:
+                    assert isinstance(response_data["product"], dict)
+                    assert isinstance(response_data["product"]["name"], str)
```
wRAR, line 126: These are needed because of how `response_data` is typed, otherwise mypy thinks the values could be anything supported by JSON.

tests/test_providers.py
```diff
@@ -198,3 +198,61 @@ async def test_provider_params_remove_unused_options(mockserver):
...
+async def test_provider_extractfrom(mockserver):
+    from typing import Annotated
+
+    @attrs.define
+    class AnnotatedProductPage(BasePage):
+        product: Annotated[Product, ExtractFrom.httpResponseBody]
+        product2: Annotated[Product, ExtractFrom.httpResponseBody]
+
+    class AnnotatedZyteAPISpider(ZyteAPISpider):
+        def parse_(self, response: DummyResponse, page: AnnotatedProductPage):  # type: ignore[override]
```
wRAR, line 213: > error: Argument 2 of "parse_" is incompatible with supertype "ZyteAPISpider"; supertype defines the argument type as "ProductPage"  [override]
> note: This violates the Liskov substitution principle
> note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides

I think this is a general problem with scrapy-poet-annotated callbacks, it will also be worse when we release typed Scrapy as `Scrapy.parse` has `Response` but we often use `DummyResponse`.

kmike: Could you please open an issue for this in scrapy-poet?

kmike (approved):

tests/test_providers.py
wRAR: I will, but I'm not sure what to do with it :)

wRAR: https://github.com/scrapinghub/scrapy-poet/issues/179

tests/test_providers.py
wRAR: Not in this case, as the exception happens in the download handler, not in a callback.

wRAR: We can check caplog though.

wRAR:
So the last question is where should we define `ExtractFrom`. Do we want to expose it in the top-level `__init__.py`? If so, we shouldn't put it in `scrapy_zyte_api.providers` as importing it needs additional deps.

Gallaecio:
What about a `scrapy_zyte_api._annotations` module?

Gallaecio (approved):

PR #151, opened 2023-11-17
Parameter mapping

### LINKED ISSUE #143: Review fingerprinting after the addition of new keys
We might not want keys like `device` or `sessionContext` to affect fingerprinting.

Maybe we need to add a test that makes sure that every new parameter added to the Zyte API spec is taken into account and mapped for the different purposes it might have in this code base, to make it harder to miss this aspect.

Gallaecio:
Resolves #143.

- `_params.py` now contains a dictionary of top-level Zyte API request params with metadata about them, and it should be extended as new parameters are added upstream in the future. Related changes:
  - `"jobPosting"` is now recognized as an extraction type (affects whether or not `httpResponseBody` and `httpResponseHeaders` are enabled by default).
  - `"GET"` is now recognized as the default value of `"httpRequestMethod"`, and stripped from request parameters if manually defined.
- Fingerprinting changes:
  - `*Options.extractFrom` is now taken into account to determine whether or not to keep the URL fragment. The fragment is no longer kept just because automatic extraction is used, `browserHtml` must also be explicitly requested as extraction source.
  - Matching `"httpRequestBody"` and `"httpRequestText"` values do not cause different fingerprints.
  - The following additional keys are ignored for the sake of fingerprints: `sessionContext`, `sessionContextParameters`, `cookieManagement`, `requestCookies`, `responseCookies`.
  - :warning: `device` remains taken into account for fingerprinting (the default behavior for any unknown key). I am not completely sure, though. If it only affects requests (and not e.g. the viewport), maybe it would be best to ignore it during fingerprinting.
  - :warning: I wonder if we should also ignore `httpResponseHeaders` (True/False), since we already take into account other outputs it can be combined with, in line how `responseCookies` is now ignored.

To do:
- [x] Tests for new scenarios.
- [x] Update after https://github.com/scrapy-plugins/scrapy-zyte-api/pull/150.


wRAR (approved):

scrapy_zyte_api/_params.py
```diff
@@ -15,22 +15,274 @@
...
+        "default": None,
+        "is_extract_type": False,
+        "requires_browser_rendering": False,
+        "changes_fingerprint": False,
+    },
+    "echoData": {
+        "default": None,
+        "is_extract_type": False,
+        "requires_browser_rendering": False,
+        "changes_fingerprint": False,
```
kmike, line 202: It seems this does change the fingerprint, because the response contains echoData as-is

scrapy_zyte_api/_params.py
```diff
@@ -15,22 +15,274 @@
...
+        "default": [],
+        "is_extract_type": False,
+        "requires_browser_rendering": False,
+        "changes_fingerprint": False,  # Treated like headers.
+    },
+    "responseCookies": {
+        "default": False,
+        "is_extract_type": False,
+        "requires_browser_rendering": False,
+        "changes_fingerprint": False,
```
kmike, line 244: It changes the output, right?

scrapy_zyte_api/_request_fingerprinter.py
```diff
@@ -36,30 +37,28 @@ def __init__(self, crawler):
...
+        def _normalize_params(self, api_params):
+            api_params["url"] = canonicalize_url(
+                api_params["url"],
+                keep_fragments=_uses_browser(api_params),
             )
 
+            if "httpRequestText" in api_params:
+                api_params["httpRequestBody"] = b64encode(
+                    api_params.pop("httpRequestText").encode()
+                ).decode()
```
kmike, line 50: should we pop httpRequestText in this case?

Gallaecio: Note that we are only popping for the purpose of fingerprinting, `api_params` is a copy of the original params.

kmike: Yes, the idea is that the request with httpRequestText and the same httpRequestBody have the same fingerprint

kmike (approved):

PR #152, opened 2023-11-23
Fail on SPM + transparent mode

Gallaecio:
I went with raising `NotConfigured` from the handlers.

However, I wonder if instead we should treat this as a fatal error and close the spider with a custom reason.

In fact, I wonder if we should do that for other `NotConfigured` errors in the handler, since some users have a hard time figuring out those kind of issues from the logs, and a non-working HTTP/HTTPS handler feels like a critical-enough issue to warrant closing the spider until it is addressed.

### DIFF
--- a/scrapy_zyte_api/handler.py
+++ b/scrapy_zyte_api/handler.py
@@ -62,6 +62,29 @@ class ScrapyZyteAPIDownloadHandler(HTTPDownloadHandler):
             raise NotConfigured(
                 "Zyte API is disabled. Set ZYTE_API_ENABLED to True to enable it."
             )
+        if settings.getbool("ZYTE_API_TRANSPARENT_MODE", False) and (
+            settings.getbool("ZYTE_SMARTPROXY_ENABLED", False)
+            or settings.getbool("CRAWLERA_ENABLED", False)
+        ):
+            spm_setting = (
+                "ZYTE_SMARTPROXY_ENABLED"
+                if settings.getbool("ZYTE_SMARTPROXY_ENABLED", False)
+                else "CRAWLERA_ENABLED"
+            )
+            raise NotConfigured(
+                f"Both ZYTE_API_TRANSPARENT_MODE and {spm_setting} are "
+                f"enabled. You should only enable one of those at the same "
+                f"time.\n"
+                f"\n"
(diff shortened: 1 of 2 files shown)


wRAR (approved):

scrapy_zyte_api/handler.py
```diff
@@ -62,6 +62,29 @@ def __init__(
             raise NotConfigured(
                 "Zyte API is disabled. Set ZYTE_API_ENABLED to True to enable it."
             )
+        if settings.getbool("ZYTE_API_TRANSPARENT_MODE", False) and (
+            settings.getbool("ZYTE_SMARTPROXY_ENABLED", False)
```
kmike, line 66: It's possible to disable SPM using spider attributes; how do we handle this case?

Gallaecio: What if we move this to a signal handler for `spider_opened`, where we can check the spider as well, and close the spider from there in this scenario?

kmike: Sounds good. I think there is also a middleware method (is_enabled?) which can be useful, but I haven't really checked it.

BurnzZ (approved):

scrapy_zyte_api/handler.py
Gallaecio: https://github.com/scrapy-plugins/scrapy-zyte-api/pull/152/commits/5acc5321d2a263090fa1f97d11b6b38780d51a75

tests/test_downloader_middleware.py
```diff
@@ -132,3 +136,73 @@ def parse(self, response):
...
+    crawler = get_crawler(NamedSpider, settings_dict=settings)
+    await crawler.crawl()
+    assert crawler.stats.get_value("finish_reason") == "plugin_conflict"
+
+
+@ensureDeferred
+async def test_spm_conflict_smartproxy_spider_attr():
+    class SPMSpider(Spider):
+        name = "spm_spider"
+        zyte_smartproxy_enabled = True
```
kmike: Do we have a test with a setting set to True, but the attribute set to False (I assume attr takes the precendence)?

kmike (approved):

PR #154, opened 2023-12-18
add request count to max requests close reason

BurnzZ:
When looking at the list of spider jobs in Scrapy Cloud, there's a column dictating the spider's `close_reason` message. Some users have raised that it's not apparently clear from this view about the max request that was set.

What do you think about including the set number of max requests to the spider `close_reason` message?

image

### DIFF
--- a/scrapy_zyte_api/_middlewares.py
+++ b/scrapy_zyte_api/_middlewares.py
@@ -108,7 +108,9 @@ class ScrapyZyteAPIDownloaderMiddleware:
         slot.delay = 0
 
         if self._max_requests_reached(downloader):
-            self._crawler.engine.close_spider(spider, "closespider_max_zapi_requests")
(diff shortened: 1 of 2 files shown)


Gallaecio:
I am not sure about it, but not against it either.

scrapy_zyte_api/_middlewares.py
```diff
@@ -108,7 +108,9 @@ def process_request(self, request, spider):
         slot.delay = 0
 
         if self._max_requests_reached(downloader):
-            self._crawler.engine.close_spider(spider, "closespider_max_zapi_requests")
+            self._crawler.engine.close_spider(
+                spider, f"closespider_max_{self._max_requests}_zapi_requests"
```
kmike, line 112: I think it's better to have the close reasons static, so that they are not unique each run. If we need any additional information, then stats look like a better place.

PR #155, opened 2023-12-22
Improve default header handling

### LINKED ISSUE #139: Support agent headers for HTTP requests
It turns our Zyte API *can* use the `User-Agent` header, and potentially other headers that we currently ignore during automatic request parameter mapping.

We need to allow specifying such headers in a way that does not cause a warning when using automatic request parameter mapping.

We might also want to expose some setting to allow picking those headers as part of automatic request parameter mapping, instead of requiring users to set `customHttpRequestHeaders` manually on requests needing them.

Gallaecio:
Make it so that default Scrapy headers (`Accept`, `Accept-Encoding`, `Accept-Language`, `User-Agent`) are not mapped by default, but can be mapped by defining them by manually defining their value (e.g. through project or higher-priority settings, or through `Request.headers`), even when that value matches their default value.

The current behavior has shown to be an issue in some real life scenarios, where the custom headers cause a ban response where not defining them would get the right response.

Resolves #139.


docs/first-steps/setup.rst
```diff
@@ -62,6 +62,9 @@ Then, set up scrapy-zyte-api integration in ``settings.py``:
     DOWNLOADER_MIDDLEWARES = {
         "scrapy_zyte_api.ScrapyZyteAPIDownloaderMiddleware": 1000,
     }
+    SPIDER_MIDDLEWARES = {
+        "scrapy_zyte_api.ScrapyZyteAPISpiderMiddleware": 100,
+    }
```
Gallaecio, line 67: I noticed it was missing from the docs.

docs/reference/settings.rst
```diff
@@ -239,7 +239,7 @@ See :ref:`retry`.
 ZYTE_API_SKIP_HEADERS
 =====================
 
-Default: ``["User-Agent"]``
+Default: ``["Cookie"]``
```
Gallaecio, line 242: `Cookie` should have been here, i.e. the documentation was out of date, the actual change was the removal of `User-Agent`, to be handled based on whether its value is used-defined or not.

Gallaecio:
Not that the initial draft is done, I have realized a potentially big issue: if the headers are defined through a custom middleware, they are still ignored :sweat:

wRAR (approved):

Gallaecio:
Once https://github.com/scrapinghub/scrapy-poet/pull/197 is merged and released, we should close and reopen this PR, and all tests should pass then.

PR #156, opened 2023-12-26
Bump scrapy-poet

### LINKED ISSUE #91: ZyteApiProvider could make an unneeded API request

wRAR:
Fixes #91 

Not sure if we should make a new release just with this.

### DIFF
--- a/docs/usage/scrapy-poet.rst
+++ b/docs/usage/scrapy-poet.rst
@@ -53,12 +53,6 @@ set the default options for various extraction types, e.g.::
         "productNavigationOptions": {"extractFrom": "httpResponseBody"},
     }
 
-Note that the built-in ``scrapy_poet.page_input_providers.ItemProvider`` has a
-priority of 2000, so when you have page objects producing
-:class:`zyte_common_items.Product` items you should use higher values for
-``ZyteApiProvider`` if you want these items to come from these page objects,
-and lower values if you want them to come from Zyte API.
-
 Dependency annotations
 ======================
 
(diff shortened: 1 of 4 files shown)


wRAR:
There is another thing we need to change here though: https://scrapy-zyte-api.readthedocs.io/en/latest/usage/scrapy-poet.html#default-parameters says "Note that the built-in scrapy_poet.page_input_providers.ItemProvider has a priority of 2000, so when you have page objects producing zyte_common_items.Product items you should use higher values for ZyteApiProvider if you want these items to come from these page objects, and lower values if you want them to come from Zyte API."

Now I think the items will always come from the page objects and if you want to get them directly you need to change the rules?

Gallaecio:
> There is another thing we need to change here though: https://scrapy-zyte-api.readthedocs.io/en/latest/usage/scrapy-poet.html#default-parameters says "Note that the built-in scrapy_poet.page_input_providers.ItemProvider has a priority of 2000, so when you have page objects producing zyte_common_items.Product items you should use higher values for ZyteApiProvider if you want these items to come from these page objects, and lower values if you want them to come from Zyte API."
> 
> Now I think the items will always come from the page objects and if you want to get them directly you need to change the rules?

I think to get them directly you should disable the corresponding rule altogether. But honestly, I think the “right” behavior is for items to come from a page object when one exists, so I think we can simply remove that paragraph altogether and not document what to do for the other behavior, at least until we find a real use case for it that we want to support (which might require code changes).

Gallaecio (approved):

BurnzZ (approved):

kmike (approved):

PR #159, opened 2024-01-05
Add JobPosting support to the provider.

### LINKED ISSUE #158: Add JobPosting to the provider
Comparing https://docs.zyte.com/zyte-api/usage/extract.html#zyte-api-extract-fields to https://scrapy-zyte-api.readthedocs.io/en/latest/reference/inputs.html#inputs I see `JobPosting` is missing, I think it should be trivial to add it or are there any problems?

wRAR:
Fixes #158

- [x] Test against the real server

### DIFF
--- a/docs/reference/inputs.rst
+++ b/docs/reference/inputs.rst
@@ -19,6 +19,8 @@ Inputs
 
 -   :class:`zyte_common_items.ArticleNavigation`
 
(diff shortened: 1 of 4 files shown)


kmike (approved):
+1 to merge after it's tested against real server

PR #161, opened 2024-01-16
support AnyResponse

BurnzZ:
A way to address https://github.com/zytedata/zyte-spider-templates/issues/25.

Related PRs:
- https://github.com/scrapinghub/web-poet/pull/195
- https://github.com/scrapinghub/scrapy-poet/pull/184
- https://github.com/zytedata/zyte-spider-templates/pull/28

# TODO:
- [x] tests
- [x] docs
- [x] Use new release of `web-poet`
- [x] Use new release of `scrapy-poet`


scrapy_zyte_api/providers.py
```diff
@@ -116,6 +122,20 @@ async def __call__(  # noqa: C901
             options_name = f"{kw}Options"
             if item_type not in to_provide_stripped and options_name in zyte_api_meta:
                 del zyte_api_meta[options_name]
```
BurnzZ, line 124: When Zyte API's Auto-Configurator would be released, we won't have the need for `web_poet.HttpOrBrowserResponse` since we can receive the corresponding extraction source from the API Response.

More or less, the same code would be used after some slight tweaks.

kmike: I think we'll still need HttpOrBrowserResponse dependency, as a way to trigger this API behavior.

scrapy_zyte_api/providers.py
```diff
@@ -116,6 +122,20 @@ async def __call__(  # noqa: C901
...
+                extract_from = zyte_api_meta[options_name]["extractFrom"]
+
+        if HttpOrBrowserResponse in to_provide:
+            if extract_from == "browserHtml":
+                html_requested = True
+            elif extract_from == "httpResponseBody":
+                param_parser = _ParamParser(crawler)
+                params = param_parser.parse(request)
+                del params["url"]
+                zyte_api_meta.update(params)
```
kmike, line 135: I didn't get this part, could you please explain the logic here?

scrapy_zyte_api/providers.py
```diff
@@ -146,6 +166,32 @@ async def __call__(  # noqa: C901
             results.append(response)
             self.update_cache(request, {BrowserResponse: response})
 
+        if HttpOrBrowserResponse in to_provide:
+            if extract_from == "browserHtml":
```
kmike, line 170: I'm not sure I understand this part. Why is it tied to extract_from? I was thinking the logic is this: if you have BrowserHtml in the output, you use it to create HttpOrBrowserResponse. If not, you use HttpResponse. It seems at this point extract_from shouldn't matter.

scrapy_zyte_api/providers.py
BurnzZ: I was thinking that it would be triggered by the new Zyte API param that would return the extraction source that was used. For example, having `{..., "anyResponse": True}` in the Zyte API request would return `httpResponseBody`+`httpResponseHeaders` or `browserHtml` in Zyte API's response.

kmike: Right, and this "anyResponse" parameter is going to be included in the API request sent by scrapy-zyte-api if a page object (or a callback) uses HttpOrBrowserResponse dependency, and won't be included otherwise, right?

BurnzZ: Aha, yes I see your point now 👍

scrapy_zyte_api/providers.py
BurnzZ: The line `params = param_parser.parse(request)` would return something lke:
```
{'customHttpRequestHeaders': [{'name': 'Accept',
                               'value': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'},
                              {'name': 'Accept-Language', 'value': 'en'}],
 'httpResponseBody': True,
 'httpResponseHeaders': True,
 'url': 'https://books.toscrape.com'}
```
which are needed for the `HttpResponse`.

scrapy_zyte_api/providers.py
BurnzZ: Good point. Updated in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/3bc8b1462d58b5039f228e345173af0a05f38113

scrapy_zyte_api/providers.py
```diff
@@ -146,6 +166,33 @@ async def __call__(  # noqa: C901
             results.append(response)
             self.update_cache(request, {BrowserResponse: response})
 
+        if AnyResponse in to_provide:
+            if "browserHtml" in api_response.raw_api_response:
+                any_response = AnyResponse(
+                    response=BrowserResponse(
```
kmike, line 172: A small thing, but probably it'd be better to reuse BrowserResponse instance if it's already created. This way we e.g. won't parse HTML twice.

BurnzZ: Good call. Added a test for it as well https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/5c00cdc76c50988a21ff1c5cd1fc83a69242c204#diff-864023bf5f9f833678b86d78cf8354385edaf4439d52876ab6b819bfe58d8c35R323-R328

BurnzZ:
Added test cases and identified new scenarios that aren't handled properly, so the CI would still break until it's updated.

tests/test_providers.py
```diff
@@ -263,3 +275,464 @@ def parse_(self, response: DummyResponse, page: AnnotatedProductPage):  # type:
...
+
+    assert type(item["page"].response) == AnyResponse
+    assert type(item["page"].response.response) == HttpResponse
+    assert type(item["page"].product) == Product
+
+
+# The issue here is that HttpResponseProvider runs earlier than ScrapyZyteAPI.
+# HttpResponseProvider doesn't know that it should not run since ScrapyZyteAPI
+# could provide HttpResponse in anycase.
+@pytest.mark.xfail(reason="Not supported yet", raises=AssertionError, strict=True)
```
BurnzZ, line 557: This is something that might need more changes and can be handled in another PR.

tests/test_providers.py
```diff
@@ -263,3 +275,464 @@ def parse_(self, response: DummyResponse, page: AnnotatedProductPage):  # type:
...
+    assert params[0].keys() == {
+        "url",
+        "httpResponseBody",
+        "customHttpRequestHeaders",
+        "httpResponseHeaders",
+    }
+    assert params[1].keys() == {"url", "browserHtml"}
+
+    assert type(item["page"].response) == AnyResponse
+    assert type(item["page"].response.response) == HttpResponse
```
BurnzZ, line 736: FIXME: Perhaps when `HttpResponse` and `BrowserResponse` are both available, `BrowserResponse` takes precedence.

BurnzZ: Done in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/85f355d3d91a6c4713746f12b9f8dbd96835405e

docs/reference/inputs.rst
```diff
@@ -13,6 +13,12 @@ Inputs
 
 -   :class:`web_poet.BrowserResponse`
 
+-   :class:`web_poet.AnyResponse`
+
+    This re-uses either :class:`web_poet.BrowserResponse` *(takes priority)*
+    or :class:`web_poet.HttpResponse` if they're available. If neither is
+    available, then it won't cause either one to be obtained from Zyte API.
```
kmike, line 20: Hm, this doesn't look like a desired behavior. I think if neither is available, we should get one of them.

BurnzZ: Would `BrowserResponse` be preferred to be obtained if neither is available?

scrapy_zyte_api/providers.py
```diff
@@ -112,10 +124,25 @@ async def __call__(  # noqa: C901
...
         for item_type, kw in item_keywords.items():
             options_name = f"{kw}Options"
             if item_type not in to_provide_stripped and options_name in zyte_api_meta:
                 del zyte_api_meta[options_name]
+            elif options_name in zyte_api_meta:
+                extract_from = zyte_api_meta[options_name]["extractFrom"]
+
+        if AnyResponse in to_provide:
+            if extract_from == "browserHtml":
+                html_requested = True
```
wRAR, line 137: Does the new logic still need this flag?

docs/reference/inputs.rst
kmike: I think it should be HttpResponse if neither is available.

BurnzZ: Thanks for confirming. Updated it in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/137d146b13d78a8c55253955183f0509f0d9fc76

scrapy_zyte_api/providers.py
BurnzZ: Currently, it's still being used. Though it might not be since I'm planning to simplify and refactor this already massive `__call__()` method we have.

wRAR (approved):

scrapy_zyte_api/providers.py
```diff
@@ -40,33 +46,42 @@ class ZyteApiProvider(PageObjectInputProvider):
...
+            item = self.injector.weak_cache.get(request, {}).get(cls)
             if item:
                 results.append(item)
                 to_provide.remove(cls)
+
+            # BrowserResponse takes precedence than HttpResponse
+            elif cls == AnyResponse and BrowserResponse not in to_provide:
+                http_response = self.injector.weak_cache.get(request, {}).get(
+                    HttpResponse
+                )
```
kmike, line 78: How do we ensure that the provider which creates HttpResponse is executed before this provider?

BurnzZ: By default, `scrapy_poet.page_input_providers.HttpResponseProvider` is given a priority of **500** ([code ref](https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/downloadermiddlewares.py#L32)), while `scrapy_zyte_api.providers.ZyteApiProvider` only has the recommendation [from the docs](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/docs/first-steps/scrapy-poet.rst) to be assigned with **1100**.

Hard to ensure as `ZyteApiProvider` isn't assigned with a priority value by default and needs manual user assignment. But if the priority configurations are set right, this case is captured in the [test_provider_any_response_browser_http_response](https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/files/932262288789a774476ca3d6474caf5d925703dc#diff-864023bf5f9f833678b86d78cf8354385edaf4439d52876ab6b819bfe58d8c35R748-R781) test case.

kmike: Aha, right. For me it was not clear that the instances are built in the order of provider priority, but indeed that's the case.

scrapy_zyte_api/providers.py
```diff
@@ -40,33 +46,42 @@ class ZyteApiProvider(PageObjectInputProvider):
...
 
+        http_response = None
         for cls in list(to_provide):
-            item = self._cached_instances.get(request, {}).get(cls)
+            item = self.injector.weak_cache.get(request, {}).get(cls)
             if item:
                 results.append(item)
                 to_provide.remove(cls)
+
+            # BrowserResponse takes precedence than HttpResponse
```
Gallaecio, line 74: :slightly_smiling_face: 

```suggestion
            # BrowserResponse takes precedence over HttpResponse
```

scrapy_zyte_api/providers.py
```diff
@@ -112,10 +125,35 @@ async def __call__(  # noqa: C901
...
+            and not http_response
+        )
+
+        extract_from = None  # type: ignore[assignment]
         for item_type, kw in item_keywords.items():
             options_name = f"{kw}Options"
             if item_type not in to_provide_stripped and options_name in zyte_api_meta:
                 del zyte_api_meta[options_name]
+            elif options_name in zyte_api_meta:
+                extract_from = zyte_api_meta[options_name]["extractFrom"]
```
Gallaecio, line 141: Do we have a guarantee that the `extractFrom` key exists in `zyte_api_meta[options_name]`, or could this raise a `KeyError`?

scrapy_zyte_api/providers.py
```diff
@@ -112,10 +125,35 @@ async def __call__(  # noqa: C901
...
+
+        if AnyResponse in to_provide:
+            if extract_from == "browserHtml":
+                html_requested = True
+            elif extract_from == "httpResponseBody" or http_response_needed:
+                param_parser = _ParamParser(crawler)
+                param_parser._transparent_mode = True
+                http_request_params = param_parser.parse(request)
+                del http_request_params["url"]
+                zyte_api_meta.update(http_request_params)
```
Gallaecio, line 153: I think we may need to use this above for browser as well, if the goal is to map things like `requestHeaders`.

I also wonder if it might be safer to run `param_parser._automap_params = zyte_api_meta` than to merge the dicts later, but I think at the moment they might be functionally equivalent

scrapy_zyte_api/providers.py
```diff
@@ -112,10 +125,35 @@ async def __call__(  # noqa: C901
...
+
+        extract_from = None  # type: ignore[assignment]
         for item_type, kw in item_keywords.items():
             options_name = f"{kw}Options"
             if item_type not in to_provide_stripped and options_name in zyte_api_meta:
                 del zyte_api_meta[options_name]
+            elif options_name in zyte_api_meta:
+                extract_from = zyte_api_meta[options_name]["extractFrom"]
+            elif item_type in to_provide_stripped and http_response_needed:
+                zyte_api_meta[options_name] = {"extractFrom": "httpResponseBody"}
```
Gallaecio, line 143: Could `"ZYTE_API_PROVIDER_PARAMS": {"productOptions": {}}` trigger a `KeyError` here?

tests/test_providers.py
Gallaecio: Should we document the disabling of the HttpResponseProvider as part of scrapy-zyte-api configuration?

BurnzZ: No, as that would prevent certain cases from being supported. For example: https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/files/932262288789a774476ca3d6474caf5d925703dc#r1464355664.

Could also be the case that `HttpResponse` comes from another source (e.g. Crawlera, though least likely).

scrapy_zyte_api/providers.py
BurnzZ: I don't think so, since before this line, the following is executed for each item type. 

```python
options = zyte_api_meta.setdefault(f"{kw}Options", {})
options["extractFrom"] = extract_from.value
```

Though I agree it'd be better to have `extract_from = zyte_api_meta[options_name].get("extractFrom")` instead, in case some future refactor changes this assumption. I'll update this. 👍

scrapy_zyte_api/providers.py
BurnzZ: I've added a TODO comment about mapping `requestHeaders` for `browserHtml` for the future. I'm worried that it might cause some regressions if introduced in this PR.

Thanks for the `param_parser._automap_params = zyte_api_meta` tip. I tried this but there's a few tests that fails if I use it. In particular, some zyte parameters will be missing.

Gallaecio (approved):

PyExplorer (approved):

scrapy_zyte_api/providers.py
```diff
@@ -112,10 +125,36 @@ async def __call__(  # noqa: C901
...
+
+        extract_from = None  # type: ignore[assignment]
         for item_type, kw in item_keywords.items():
             options_name = f"{kw}Options"
             if item_type not in to_provide_stripped and options_name in zyte_api_meta:
                 del zyte_api_meta[options_name]
+            elif options_name in zyte_api_meta:
+                extract_from = zyte_api_meta[options_name].get("extractFrom")
+            elif item_type in to_provide_stripped and http_response_needed:
+                zyte_api_meta[options_name] = {"extractFrom": "httpResponseBody"}
```
kmike, line 143: To check my understanding: the logic here is that if browserHtml is not requested, but a data type is requested, and there is AnyResponse, then we switch the extractFrom from default to httpResponseBody for this data type?

scrapy_zyte_api/providers.py
```diff
@@ -112,10 +125,36 @@ async def __call__(  # noqa: C901
...
+            and not http_response
+        )
+
+        extract_from = None  # type: ignore[assignment]
         for item_type, kw in item_keywords.items():
             options_name = f"{kw}Options"
             if item_type not in to_provide_stripped and options_name in zyte_api_meta:
                 del zyte_api_meta[options_name]
+            elif options_name in zyte_api_meta:
+                extract_from = zyte_api_meta[options_name].get("extractFrom")
```
kmike, line 141: I think currently it's expected to work fine, but if we add any other option to the API, then extract_from could become None, and it seems this case is not handled later. 

Would it make sense to change the logic so that the condition below is still executed if extractFrom is None after this line?

scrapy_zyte_api/providers.py
```diff
@@ -112,10 +125,36 @@ async def __call__(  # noqa: C901
...
+
+        if AnyResponse in to_provide:
+            if extract_from == "browserHtml":
+                html_requested = True
+            elif extract_from == "httpResponseBody" or http_response_needed:
+                param_parser = _ParamParser(crawler)
+                param_parser._transparent_mode = True
+                http_request_params = param_parser.parse(request)
+                del http_request_params["url"]
+                zyte_api_meta.update(http_request_params)
```
kmike, line 153: TBH, this logic is the least clear to me. Could you please add a comment, to explain a bit how it works?

1. In the docs we write that `scrapy-poet integration ignores default parameters`. But it seems here they are applied, and ZYTE_API_PROVIDER_PARAMS are ignored instead? Or is it not happening because of some reason?
2. We parse the original request to get the additional keywords to add to zyte_api_meta. I wonder how it works in cases where the original request contains zyte_api_meta itself, and if behavior is different from how the provider works usually in such cases.
3. `del http_request_params["url"]` is mysterious to me :) Why delete the url? Are there other parameters which need to be deleted?

kmike: Is the idea here is to handle cookies, headers, etc. in a more consistent way, as compared to just setting httpRequestBody and httpRequestHeaders to True, without invoking ParamParser? 

What are the actual differences? What breaks if ParamParser is not used?

scrapy_zyte_api/providers.py
BurnzZ: It'd be, if:
- there's no explicit extraction source requested for the given `item_type` (e.g. Product, ProductNavigation, etc), and
- AnyResponse is one of the requested, and
- either BrowserResponse or BrowserHtml is not requested, and
- HttpResponse has not been created by previous providers (i.e. HttpResponseProvider)

then we use `httpResponseBody` as the extraction source.

scrapy_zyte_api/providers.py
BurnzZ: Good call. Updated in https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/3fec1b8fa5e06e876da14ac5a494970bd8b8bcf3 and added tests for this

scrapy_zyte_api/providers.py
BurnzZ: > In the docs we write that scrapy-poet integration ignores default parameters. But it seems here they are applied, and ZYTE_API_PROVIDER_PARAMS are ignored instead? Or is it not happening because of some reason?

That's a good point. I forgot about this and so using the `ParamUser` was a way to make handling the headers consistent across the requests. I can't say for the actual differences in practice. For now, we can go with the simplest approach of setting `httpResponseBody` and `httpResponseHeaders` to True. https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/b3419765e0429d7bebe971e0bdff73ca99932996

kmike: @BurnzZ simply setting httpResponseBody/Headers looks way easier to understand, as it's similar to how all other parameters are handled. 

@Gallaecio what do you think about this? Do you see any edge cases with using ParamParser vs just requesting a httpResponseBody/Headers? Any concerns about using the parameters directly, without ParamParser?

Gallaecio: The only case I can think of, and would have been a problem already with the pre-existing code, is the case where a cookie included in the request is necessary to get the right content, or some actions are necessary for extraction to work properly, and the server side cannot (yet) inject those automatically.

Still, it may be better to keep things simple for now, and figure out how we wish to solve these issues when we get to that. Because even if we decide to use ParamParser, things are more complicated: it should only be used if automatic parameter parsing is being used, if (raw) zyte_api is used for the source request then some parameters may also need to be copied from there…

BurnzZ: Good point about the necessary cookies. Fortunately, I haven't encountered this yet in my experiments.

+1 on keeping things simple for now.

scrapy_zyte_api/providers.py
```diff
@@ -40,34 +45,43 @@ class ZyteApiProvider(PageObjectInputProvider):
...
     def is_provided(self, type_: Callable) -> bool:
         return super().is_provided(strip_annotated(type_))
 
     def update_cache(self, request: Request, mapping: Dict[Any, Any]) -> None:
-        if request not in self._cached_instances:
-            self._cached_instances[request] = {}
-        self._cached_instances[request].update(mapping)
+        if request not in self.injector.weak_cache:
+            self.injector.weak_cache[request] = {}
+        self.injector.weak_cache[request].update(mapping)
```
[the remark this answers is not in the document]
BurnzZ: Currently, updating the cache doesn't do anything, unless there's another provider that runs much later that would need it.

In the case for `ZyteApiProvider`, it reads the entries of `self.injector.weak_cache` to get the **HttpResponse** instance from the `HttpResponseProvider` that runs much earlier.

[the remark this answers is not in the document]
kmike: ah, right, if we have HttpResponse in cache, other providers should also be adding things to cache, even if we're not using it now.

Should we move this update_cache (or a similar method) to Injector?

tests/test_providers.py
```diff
@@ -312,3 +324,632 @@ def parse_(self, response: DummyResponse, page: GeoProductPage):  # type: ignore
...
+    )
+    assert len(results) == 1
+    assert type(results[0]) == AnyResponse
+    assert type(results[0].response) == HttpResponse
+
+    # Same case as above, HttpResponse is used by default
+    results = yield provide({AnyResponse, Product})
+    assert len(results) == 2
+    assert type(results[0]) == AnyResponse
+    assert type(results[1]) == Product
```
kmike, line 374: Should there be an assert for `type(results[0].response) == HttpResponse`? Or should it be BrowserResponse here (i.e. the comment above is outdated)?

tests/test_providers.py
```diff
@@ -312,3 +324,632 @@ def parse_(self, response: DummyResponse, page: GeoProductPage):  # type: ignore
...
+
+    # AnyResponse should re-use BrowserResponse if available.
+    results = yield provide({AnyResponse, BrowserResponse})
+    assert len(results) == 2
+    assert type(results[0]) == BrowserResponse
+    assert type(results[1]) == AnyResponse
+    assert id(results[0]) == id(results[1].response)
```
kmike, line 381: Would this work?

```suggestion
    assert results[1].response is results[0]
```

tests/test_providers.py
```diff
@@ -312,3 +324,632 @@ def parse_(self, response: DummyResponse, page: GeoProductPage):  # type: ignore
...
+    }
+
+    assert type(item["page"].response) == AnyResponse
+    assert type(item["page"].response.response) == BrowserResponse
+    assert type(item["page"].product) == Product
+
+
+# The issue here is that HttpResponseProvider runs earlier than ScrapyZyteAPI.
+# HttpResponseProvider doesn't know that it should not run since ScrapyZyteAPI
+# could provide HttpResponse in anycase.
```
kmike, line 697: What is the failure? Sorry, I haven't checked the logs :)

kmike (approved):
I've added a few comments, but it looks good overall @BurnzZ - great work! + to merge after updating to the released scrapy-poet.

tests/test_providers.py
BurnzZ: The issue happens if `HttpResponse` is explicitly declared as a dependency, say in the PO.

Since `HttpResponseProvider` runs much earlier than `ZyteApiProvider`, it would make a request to ZAPI. When it's `ZyteApiProvider` turn to fullfill dependencies, it would have a 2nd request to ZAPI to fulfill the `AnyResponse` + `Product` dependencies.

The ideal scenario would be having only a single ZAPI request that would fulfill all three of the `HttpResponse`, `AnyResponse`, and `Product` dependencies.

It could be the case that in another PR, we can combine `HttpResponseProvider` and `ZyteApiProvider` together, or perhaps create a decision mechanism to determine which providers should run that would result in a more optimal dependency creation.

I'm not sure how often this would occur in practice though, since if you have `AnyResponse`, there's not much need to declare an `HttpResponse` dependency. With that, it should be easy to avoid.

tests/test_providers.py
BurnzZ: I've removed this test since they duplicate the other test cases that follow it. Not to mention these also lack certain cases. https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/df32f14598238a0166e8ba488c45d1857e497e55

tests/test_providers.py
```diff
@@ -312,3 +324,632 @@ def parse_(self, response: DummyResponse, page: GeoProductPage):  # type: ignore
...
+    results = yield provide({AnyResponse, Product})
+    assert len(results) == 2
+    assert type(results[0]) == AnyResponse
+    assert type(results[1]) == Product
+
+    # AnyResponse should re-use BrowserResponse if available.
+    results = yield provide({AnyResponse, BrowserResponse})
+    assert len(results) == 2
+    assert type(results[0]) == BrowserResponse
+    assert type(results[1]) == AnyResponse
```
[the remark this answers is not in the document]
BurnzZ: Similar to https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161#discussion_r1482507898, I've removed these tests Mikhail since they duplicate the newer tests below, which are far more complete. These tests also don't cover how they interact with the other providers _(e.g. sharing instances via the `Injector.weak_cache`)_ so this type of test implementation is far too limited.

PR #162, opened 2024-01-16
Apply request slotting also on the spider middleware

Gallaecio:


PyExplorer:
With ScrapyZyteAPIDownloaderMiddleware the slot names are not changed now - I've checked with the Articles+DownloaderAwarePriorityQueue - all good. 
Thanks @Gallaecio! Please let me know when it will be merged and ready to use.

scrapy_zyte_api/_middlewares.py
```diff
@@ -99,15 +98,7 @@ def process_request(self, request, spider):
         if self._param_parser.parse(request) is None:
             return
 
-        downloader = self._crawler.engine.downloader
-        slot_id = downloader._get_slot_key(request, spider)
-        if not isinstance(slot_id, str) or not slot_id.startswith(self._slot_prefix):
-            slot_id = f"{self._slot_prefix}{slot_id}"
-            request.meta["download_slot"] = slot_id
-        _, slot = downloader._get_slot(request, spider)
-        slot.delay = 0
```
kmike, line 108: So, setting slot in the download middleware is not needed anymore?

Gallaecio: Right, we should keep the code to support requests that come from downloader middlewares. For such requests the queue issue would persist, since nothing will be able to set their slot before they reach the scheduler, but they would not be affected by the AutoThrottle extension at least.

scrapy_zyte_api/_middlewares.py
```diff
@@ -13,18 +13,35 @@
...
+            slot_id = f"{self._slot_prefix}{slot_id}"
+            request.meta["download_slot"] = slot_id
+        _, slot = downloader._get_slot(request, spider)
+        slot.delay = 0
+
+
+class ScrapyZyteAPIDownloaderMiddleware(_BaseMiddleware):
     @classmethod
     def from_crawler(cls, crawler):
         return cls(crawler)
```
kmike, line 39: I think you can move this method to the base class as well

BurnzZ (approved):

PR #167, opened 2024-01-29
Add a Geolocation dependency.

wRAR:


docs/usage/scrapy-poet.rst
```diff
@@ -72,3 +72,21 @@ supported one is :class:`scrapy_zyte_api.ExtractFrom`:
...
+.. code-block:: python
+
+    from typing import Annotated
+
+    from scrapy_zyte_api import Geolocation
+
+    @attrs.define
+    class MyPageObject(BasePage):
+        product: Product
+        geolocation: Annotated[Geolocation, "DE"]
```
kmike, line 92: I haven't checked it, but onder if something like this would work:

```py
class _Geolocation:
    pass

def Geolocation(geolocation: str):
    return Annotated[_Geolocation, geolocation]

@attrs.define
class MyPageObject(ProductPage):
    geolocation: Geolocation("DE")
```

Where does it break?

wRAR: `error: Invalid type comment or annotation  [valid-type]` from mypy, though the runtime code works.

scrapy_zyte_api/providers.py
```diff
@@ -91,26 +92,32 @@ async def __call__(  # noqa: C901
...
-            kw = item_keywords.get(cls_stripped)
-            if not kw:
-                continue
+            if cls_stripped is not Geolocation:
+                kw = item_keywords.get(cls_stripped)
+                if not kw:
+                    continue
+                zyte_api_meta[kw] = True
+            else:
+                kw = None
```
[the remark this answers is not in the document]
wRAR: Perfect!

docs/usage/scrapy-poet.rst
kmike: I see. Let's not handle it here then. This mypy issue seems relevant: https://github.com/python/mypy/issues/13643 (with a suggested workaround).

scrapy_zyte_api/providers.py
```diff
@@ -91,6 +92,9 @@ async def __call__(  # noqa: C901
         for cls in to_provide:
             cls_stripped = strip_annotated(cls)
             assert isinstance(cls_stripped, type)
+            if cls_stripped is Geolocation and is_typing_annotated(cls):
+                zyte_api_meta["geolocation"] = cls.__metadata__[0]  # type: ignore[attr-defined]
```
kmike, line 96: Should we raise an error / warning if it's not annotated?

scrapy_zyte_api/providers.py
```diff
@@ -150,6 +154,10 @@ async def __call__(  # noqa: C901
             cls_stripped = strip_annotated(cls)
             assert isinstance(cls_stripped, type)
             kw = item_keywords.get(cls_stripped)
```
kmike, line 156: Should this kw = ... be moved belog geolocation handling code?

scrapy_zyte_api/providers.py
```diff
@@ -150,6 +154,10 @@ async def __call__(  # noqa: C901
...
+            if cls_stripped is Geolocation and is_typing_annotated(cls):
+                item = AnnotatedResult(Geolocation(), cls.__metadata__)  # type: ignore[attr-defined]
+                results.append(item)
```
kmike, line 159: Is there a need to update cache?

wRAR: As I understand it the purpose of the cache is to skip requesting things, and we don't request Geolocation, but also we don't want to remove it from to_provide like it would be done if it's found in the cache. On the other hand  I don't know in which cases is the cache currently used.

wRAR: Also is it possible to request a Product without a Geolocation, cache it and then get that instance when the provider code is called again with a Geolocation, thus getting a wrong one?

PR #172, opened 2024-02-05
Remove the cache from the provider.

wRAR:
This removes the cache added in #94. It's possible that there are use cases that I don't see (and which aren't covered by tests) where it's used, but it looks like we shouldn't call the provider multiple times for the same request?

### DIFF
--- a/scrapy_zyte_api/providers.py
+++ b/scrapy_zyte_api/providers.py
@@ -1,5 +1,4 @@
 from typing import Any, Callable, Dict, List, Sequence, Set
-from weakref import WeakKeyDictionary
 
 from andi.typeutils import is_typing_annotated, strip_annotated
 from scrapy import Request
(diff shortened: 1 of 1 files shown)


Gallaecio:
I imagine this was necessary before your item building refactoring, where resolution happened at different stages and the same provider could be called multiple times, and it is not needed anymore. But I suspect @BurnzZ might have a better view here (or/and I might be mixing different caches in my mind).

BurnzZ:
This coincides with https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/ where we still kinda need the cache, albeit replaced with the new one in https://github.com/scrapinghub/scrapy-poet/pull/184.

BurnzZ:
Ahh I think the lines like `self.update_cache(request, {BrowserHtml: html})` aren't needed anymore. Perhaps it's best to have this after https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/ is merged?

wRAR:
Do we only need the scrapy-poet one then?

kmike:
Is this still valid?

wRAR:
Done in #161

PR #178, opened 2024-03-11
Support providing screenshots

Gallaecio:


scrapy_zyte_api/providers.py
```diff
@@ -183,6 +194,11 @@ async def __call__(  # noqa: C901
             )
             results.append(browser_response)
 
+        if screenshot_requested:
+            screenshot_b64 = api_response.raw_api_response["screenshot"]
+            screenshot = Screenshot(body=b64decode(screenshot_b64.encode()))
```
kmike, line 199: What do you think about moving this code to Screenshot itself? E.g. from_base64 classmethod, or maybe even some converter for the field (though probably not)?

scrapy_zyte_api/__init__.py
```diff
@@ -13,3 +13,10 @@
 from ._request_fingerprinter import ScrapyZyteAPIRequestFingerprinter
 from .addon import Addon
 from .handler import ScrapyZyteAPIDownloadHandler
+
+try:
+    import web_poet
```
kmike, line 18: it seems this is not needed anymore

tests/test_providers.py
```diff
@@ -839,3 +846,39 @@ def parse_(self, response: DummyResponse, page1: FirstPage, page2: SecondPage):
...
     assert type(item["page2"].response) is AnyResponse
     assert type(item["page2"].response.response) is BrowserResponse
+
+
+@ensureDeferred
+async def test_screenshot():
+    settings = create_scrapy_settings()
+    settings["SCRAPY_POET_PROVIDERS"] = {ZyteApiProvider: 0}
+    crawler = get_crawler(settings_dict=settings)
+    injection_mw = InjectionMiddleware.from_crawler(crawler)
```
kmike, line 856: Would it make sense to test this feature using mockserver instead? I.e. add screenshot feature to the mockserver, and then just check that the dependency works as expected?

scrapy_zyte_api/page_inputs.py
```diff
@@ -0,0 +1,9 @@
+import attrs
+from web_poet import Injectable
+
+
+@attrs.define
+class Screenshot(Injectable):
```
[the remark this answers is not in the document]
kmike: I think that's fine, but I'd prefer to rename it now. Screenshot has nothing to do with annotations (i.e. Annotated support).

tests/test_providers.py
Gallaecio: I made an effort to avoid it due to https://github.com/scrapy/scrapy/issues/6255. I also feel like off-mockserver approaches give more granular control, although they tend to be less readable and more likely to break with future upstream API changes.

Shall I switch it to mockserver nonetheless?

kmike: Mockserver doesn't have to be slow: the startup time can be amortized using pytest fixtures bound to session, and other things should be pretty fast. I recall it was quite fast e.g. in Splash. Haven't checked the Scrapy issue.

Probably that's personal, but I find low-level mock tests to often be an obstacle (instead of a help) in case of refactoring. Also, they don't actually show that the feature works from the user point of view.

kmike:
Looks good, thanks @Gallaecio!

PR #185, opened 2024-04-05
python-zyte-api 0.5.0 support

Gallaecio:


Gallaecio:
Some tests are hanging with the new API. I suspect it has to do not with the new API, but with making engine_started async. But I’m having a hard time figuring it out, so I’ll come back later with fresh eyes (if anyone wants to give it a go before that, be my guess).

scrapy_zyte_api/handler.py
```diff
@@ -100,12 +95,16 @@ def __init__(
...
 
     @classmethod
     def from_crawler(cls, crawler):
         return cls(crawler.settings, crawler)
 
-    def engine_started(self):
+    async def engine_started(self):
+        self._session = await self._client.session(
+            trust_env=self._trust_env
+        ).__aenter__()
```
kmike, line 107: I'm not sure `__aenter__` is needed here. aiohttp.ClientSession example don't always use it (so, it seems not using a context manager is ok), and `__aenter__` is no-op, which just returns self.

scrapy_zyte_api/handler.py
```diff
@@ -275,12 +270,12 @@ def close(self) -> Generator:
         yield deferred_from_coro(self._close())
 
     async def _close(self) -> None:  # NOQA
-        await self._session.close()
+        await self._session.__aexit__(None, None, None)
```
kmike, line 273: I think we should add close method to Session, this looks a bit ugly :)

scrapy_zyte_api/handler.py
kmike: From https://docs.aiohttp.org/en/stable/client_quickstart.html:

> A session context manager usage is not mandatory but await session.close() method should be called in this case, e.g.:

kmike: Hm, maybe it would also mean that there is no need for async def engine_started?

Gallaecio: > Hm, maybe it would also mean that there is no need for async def engine_started?

We might want to do that anyway nonetheless, to avoid [this warning](https://github.com/aio-libs/aiohttp/pull/1468).

kmike (approved):

Gallaecio:
[Test failures are unrelated to this change.](https://github.com/scrapy-plugins/scrapy-zyte-api/issues/190)

Issue #187, opened 2024-04-09
Using browser actions raises a TypeError


fcanobrash:
Hi team.

Following the [example to use Browser Actions](https://scrapy-zyte-api.readthedocs.io/en/latest/usage/scrapy-poet.html#browser-actions), we're getting the following error:

`File ".../lib/python3.11/site-packages/scrapy_zyte_api/_annotations.py", line 62, in <genexpr>
    return tuple(frozenset(action.items()) for action in value)
TypeError: unhashable type: 'dict'`

We're using the same code from the documentation:

```
from typing import Annotated

from scrapy_zyte_api import Actions, actions


@attrs.define
class MyPageObject(BasePage):
    product: Product
    actions: Annotated[
        Actions,
        actions(
            [
                {
                    "action": "click",
                    "selector": {"type": "css", "value": "button#openDescription"},
                    "delay": 0,
                    "button": "left",
                    "onError": "return",
                },
                {
                    "action": "waitForTimeout",
                    "timeout": 5,
                    "onError": "return"
                },
            ]
        ),
    ]
```

Gallaecio:
https://github.com/scrapy-plugins/scrapy-zyte-api/pull/188

PR #188, opened 2024-04-15
adding handling of nested dict for actions

emarondan:
A possible solution for #187 

- A new function `make_hashable(obj)` was added to convert objects into hashable versions.
- The `actions()` function was updated to use `make_hashable(obj)`.
- A loop was added to iterate over "actions" in `zyte_api_meta`. This allows to convert immutable frozenset objects into mutable dict objects within the "actions" of `zyte_api_meta`.

### DIFF
--- a/scrapy_zyte_api/providers.py
+++ b/scrapy_zyte_api/providers.py
@@ -119,16 +119,18 @@ class ZyteApiProvider(PageObjectInputProvider):
                         "Actions dependencies must be annotated, "
                         "e.g. Annotated[Actions, actions([...list of actions...])]."
                     )
-                zyte_api_meta["actions"] = [
-                    {
-                        k: (
-                            dict(v)
-                            if isinstance(v, frozenset)
-                            else list(v) if isinstance(v, tuple) else v
-                        )
-                        for k, v in cls.__metadata__[0][0]  # type: ignore[attr-defined]
-                    }
-                ]
+                zyte_api_meta["actions"] = []
+                for action in cls.__metadata__[0]:  # type: ignore[attr-defined]
+                    zyte_api_meta["actions"].append(
+                        {
+                            k: (
+                                dict(v)
+                                if isinstance(v, frozenset)
+                                else list(v) if isinstance(v, tuple) else v
+                            )
+                            for k, v in action
+                        }
+                    )
                 continue
             kw = item_keywords.get(cls_stripped)
             if not kw:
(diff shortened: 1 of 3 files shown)


scrapy_zyte_api/providers.py
```diff
@@ -120,6 +120,10 @@ async def __call__(  # noqa: C901
                         "e.g. Annotated[Actions, actions([...list of actions...])]."
                     )
                 zyte_api_meta["actions"] = [dict(action) for action in cls.__metadata__[0]]  # type: ignore[attr-defined]
```
wRAR, line 122: This is the line that makes the reverse conversion, it needs to be updated to be recursive too.

scrapy_zyte_api/providers.py
```diff
@@ -119,7 +119,12 @@ async def __call__(  # noqa: C901
                         "Actions dependencies must be annotated, "
                         "e.g. Annotated[Actions, actions([...list of actions...])]."
                     )
-                zyte_api_meta["actions"] = [dict(action) for action in cls.__metadata__[0]]  # type: ignore[attr-defined]
+                zyte_api_meta["actions"] = [
+                    {
+                        k: dict(v) if isinstance(v, frozenset) else v
```
wRAR, line 124: Oh, I think I didn't notice the added code is about the reverse conversion :)

I think this doesn't change tuples back to lists? I would expect dicts and lists in the result as it's a wrapper for a JSON API.

wRAR:
This needs a new test, adding a `"selector": {"type": "css", "value": "button#openDescription"}` item to the action in the existing one may be enough.

scrapy_zyte_api/providers.py
emarondan: The final result here would be a List of dicts. Do we need something different?

wRAR: Dict values can be lists, e.g. in the `values` key (looking at `Action` fields, which should match the API).

emarondan: Added a check for tuples on https://github.com/scrapy-plugins/scrapy-zyte-api/pull/188/commits/213182dd77ccfd7df4126040e3979d575db82df6

Gallaecio:
[Test failures are unrelated to this change.](https://github.com/scrapy-plugins/scrapy-zyte-api/issues/190), but the provider job failure is relevant.

emarondan:
Sorry.. forgot to add `# type: ignore[attr-defined]` comment when using `cls.__metadata__[0]` 
Done in https://github.com/scrapy-plugins/scrapy-zyte-api/commit/cba02bb7c1524b74ddf1b1f329c9b4b07ff9190b

kmike (approved):

wRAR (approved):
Thanks!

PR #202, opened 2024-06-11
Track in stats which fields from Zyte API automatic extraction are not overridden

Gallaecio:
Requested by @VMRuiz.

It adds stats that indicate which fields from dependency injection come from Zyte API automatic extraction.

When directly requesting an item type, e.g. `Product`, the entire list of fields is tracked in stats:

```
 'scrapy-zyte-api/auto_fields/zyte_common_items.items.product.Product': 'additionalProperties '
                                                                        'aggregateRating '
                                                                        'availability '
                                                                        'brand '
                                                                        'breadcrumbs '
                                                                        'canonicalUrl '
                                                                        'color '
                                                                        'currency '
                                                                        'currencyRaw '
                                                                        'description '
                                                                        'descriptionHtml '
                                                                        'features '
                                                                        'gtin '
                                                                        'images '
                                                                        'mainImage '
                                                                        'metadata '
                                                                        'mpn '
                                                                        'name '
                                                                        'price '
                                                                        'productId '
                                                                        'regularPrice '
                                                                        'size '
                                                                        'sku '
                                                                        'style '
                                                                        'url '
                                                                        'variants',
```

When using a page object that overrides some fields, the stat reflects the corresponding page object, and the override fields are removed, e.g. `aggregateRating` here:

```
 'scrapy-zyte-api/auto_fields/zyte_spider_templates_project.pages.books_toscrape_com.BooksToScrapeComProductPage': 'additionalProperties '
                                                                                                                   'availability '
                                                                                                                   'brand '
                                                                                                                   'breadcrumbs '
                                                                                                                   'canonicalUrl '
                                                                                                                   'color '
                                                                                                                   'currency '
                                                                                                                   'currencyRaw '
                                                                                                                   'description '
                                                                                                                   'descriptionHtml '
                                                                                                                   'features '
                                                                                                                   'gtin '
                                                                                                                   'images '
                                                                                                                   'mainImage '
                                                                                                                   'metadata '
                                                                                                                   'mpn '
                                                                                                                   'name '
                                                                                                                   'price '
                                                                                                                   'productId '
                                                                                                                   'regularPrice '
                                                                                                                   'size '
                                                                                                                   'sku '
                                                                                                                   'style '
                                                                                                                   'url '
                                                                                                                   'variants',
```


scrapy_zyte_api/providers.py
```diff
@@ -66,6 +151,7 @@ async def __call__(  # noqa: C901
         http_response = None
         screenshot_requested = Screenshot in to_provide
         for cls in list(to_provide):
+            self._track_auto_fields(crawler, request, cast(type, cls))
```
VMRuiz, line 154: Maybe you can add a setting to disable this in case it becomes an issue for some people?

docs/reference/settings.rst
```diff
@@ -6,6 +6,38 @@ Settings
...
+Enables stats that indicate which requested fields come directly from
+:ref:`zyte-api-extract`.
+
+If for any request no page object class is used to override
+:ref:`zyte-api-extract` fields for a given item type, the following stat is
+set:
+
+.. code-block:: python
+
+    "scrapy-zyte-api/auto_fields/<item class import path>": "<space-separated field list>"
```
kmike, line 25: How would this stat work if there are several page objects for the same item type, but for different websites / paths?

Gallaecio: This stat only happens when there is no page object for a given website, and instead the item is directly provided by Zyte API.

If there is a page object, the other stat is used instead.

docs/reference/settings.rst
```diff
@@ -6,6 +6,38 @@ Settings
...
+.. code-block:: python
+
+    "scrapy-zyte-api/auto_fields/<override class import path>": (
+        "<space-separated list of fields not overridden>"
+    )
+
+.. note:: If that page object class is not a subclass of an ``Auto``-prefixed
+    class from :doc:`zyte-common-items <zyte-common-items:index>`, all fields
+    are assumed to have been overridden, i.e. the stat value is always an empty
+    string.
```
kmike, line 39: Overall, ❤️ this feature! Nice idea @VMRuiz.

I'd like to discuss an alternative implementation, as usual :) What do you think about modifying zyte-common-items auto page classes to set custom field meta - e.g.:

```py
@attrs.define
class AutoArticlePage(BaseArticlePage):
    article: Article

    @field(meta={"auto_field": True})  # TODO: make @auto_field shortcut
    def headline(self) -> Optional[str]:
        return self.article.headline

    @field(meta={"auto_field": True})
    def datePublished(self) -> Optional[str]:
        return self.article.datePublished
```

Then, instead of using mro, etc., just access the field meta. I assume that if you override the field, you use just `@field`, so there is no meta, and we know the field is overritten. 

Also, if you want to override the field, but still indicate that it still doesn't change the data, you can add this meta to a field.

A disadvantage is that it'd require a recent zyte-common-items version to work, and that it's a bit more extensive change.

CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Changes
 =======
 
+N.N.N (YYYY-MM-DD)
+------------------
+
+* ``scrapy-zyte-api[provider]`` now requires zyte-common-items >= 0.20.0.
```
kmike, line 7: I wonder how big is it an issue, and if it worths it to have the feature disabled for older zyte-common-items versions, instead of raising an exception. @VMRuiz any opinion on that?

docs/reference/settings.rst
VMRuiz: I guess it works as a way to know if Zyte API Extraction was used. Is there any other stats to indicate that Zyte API Extraction was used without a Page Object?

Gallaecio: There’s stats for Zyte API fields used in requests, e.g.:

```
scrapy-zyte-api/request_args/browserHtml	2
scrapy-zyte-api/request_args/jobId	22
scrapy-zyte-api/request_args/product	20
scrapy-zyte-api/request_args/productNavigation	2
scrapy-zyte-api/request_args/url	22
```

CHANGES.rst
Gallaecio: > What would be the reason for this not to be compatible with older `zyte-common-items` packages?

Mainly the new `is_auto_field` method we added in 0.20.0.

Although before that I also had increased the version, to 0.19.0, I think because of the imports that the feature needed, of all the `Auto` classes.

We could make it so that the stat only works on zyte-common-items >= 0.20.0, but earlier versions can be installed, would that work?

If not, we can try to support earlier versions as well for the stat, it simply complicates the implementation a bit.

docs/reference/settings.rst
kmike: > I guess it works as a way to know if Zyte API Extraction was used. Is there any other stats to indicate that Zyte API Extraction was used without a Page Object?

What does it mean exactly, by the way? Is it "callback has an item dependency, and to resolve this dependency scrapy-zyte-api provider was used directly, witthout page objects involved"?  I wonder if we can add a related stat to scrapy-poet (not here).

kmike: If the goal is to see which items are not overwritten, it sounds easier to not have the field lists in the stat value, and have simple item:value counters instead.

CHANGES.rst
kmike: I was worried that some data types changed, but so far, based on the changelog, it seems zyte-common-items don't have any breaking changes, or at least nothing super-visible. But let's wait for @VMRuiz 's feedback, his opinion matters most here.

asadurski: From my side, I haven't noticed any conflicts. I checked requirements in the first projects we would be using this and they match.
(No functional testing performed though).

docs/reference/settings.rst
asadurski: For my side of functionality (find overwritten fields) item counts are irrelevant, but I could see some use for this when digging deeper into multi-item or multi-domain jobs.

Gallaecio: @asadurski My main question for you would be:
- Whether or not we should log in stats something at all in cases where no override is used. I imagine it might be useful in multi-domain spiders, to tell when both overrides and non-overrides were used.
- If we do log it in stats, if there is any point in listing its fields (i.e. all item fields from Zyte API), or if instead we should hard-code some short string (e.g. `(all fields)`) instead.

Gallaecio: Talked to @asadurski, agreed on keeping item entries when there are no overrides, and using a hard-coded `(all fields)` string as value.

CHANGES.rst
Gallaecio: @kmike Should we go ahead with 0.20.0 as minimum, or lower it just to be safe?

kmike: Let's proceed with 0.20.0; we can add support for older versions later if that turn out to be an issue in practice.

Gallaecio:
@asadurski All good on your end? Shall we merge and release?

asadurski (approved):

asadurski:
> @asadurski All good on your end? Shall we merge and release?

Yes, this has all the things I need and looks great!

PR #205, opened 2024-06-21
Copy known zyte_api_session-prefixed meta into session initialization requests

Gallaecio:
@VMRuiz reported that, to implement a check method that takes into account meta-defined locations, the init request also needs to have that meta key set. ~I went a bit further and made it so that any `zyte_api_session_`-prefixed meta key is copied into the init request, for extra flexibility.~

To do:
- [x] ~Docs about the meta key copy.~
- [ ] ~Docs about the need, in these scenarios, to create separate pools per location. Or maybe we should provide a default implementation that does so? Maybe even the ability to set a custom pool through meta?~ ([separate PR](https://github.com/scrapy-plugins/scrapy-zyte-api/pull/205#issuecomment-2188283970))

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     hooks:
     - id: black
 -   repo: https://github.com/pycqa/flake8
-    rev: 7.0.0
+    rev: 7.1.0
     hooks:
     - id: flake8
       additional_dependencies:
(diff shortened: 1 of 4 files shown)


Gallaecio:
I’ll keep this PR simple and easy to merge, just copy the 2 known meta keys, and open a separate PR with additional ideas brought up with working on this.

wRAR (approved):

PR #213, opened 2024-08-19
Handle custom attributes received in the API response.

wRAR:


wRAR:
Requires https://github.com/zytedata/zyte-common-items/pull/100

wRAR:
(some test failures related to field overrides)

Gallaecio (approved):

Gallaecio:
I’ll handle the update of https://github.com/scrapy-plugins/scrapy-zyte-api/blob/a1d81d11854b420248f38e7db49c685a8d46d943/scrapy_zyte_api/_params.py#L25 since I am already refactoring it significantly as part of SERP work.

scrapy_zyte_api/providers.py
```diff
@@ -186,6 +191,17 @@ async def __call__(  # noqa: C901
                         }
                     )
                 continue
+            if cls_stripped in {CustomAttributes, CustomAttributesValues}:
+                zyte_api_meta["customAttributes"] = {
+                    k: (
+                        dict(v)
+                        if isinstance(v, frozenset)
+                        else list(v) if isinstance(v, tuple) else v
+                    )
```
kmike, line 200: this code seems to be copy-pasted - time to add a helper function? Is it something like undo_make_hashable?

wRAR: Yeah.

wRAR: Done, and now I think both functions need tests. We don't actually test what we receive in mockserver requests.

kmike: Makes sense, let's add a few tests!

kmike (approved):
The code looks good. +1 to merge after CI is fixed, and some very basic tests are added for the hash/unhash functions.

wRAR:
We should release zyte-common-items before merging this I think.

kmike:
@wRAR feel free to merge when you think it's ready.

PR #227, opened 2024-10-18
Adding DOWNLOAD_MAXSIZE and DOWNLOAD_WARNSIZE

PyExplorer:
The initial version without fixing tests. The docs is still under fixing.


docs/reference/settings.rst
```diff
@@ -206,6 +206,42 @@ when :setting:`ZYTE_API_LOG_REQUESTS` is enabled, excluding object keys.
 To disable truncation, set this to ``0``.
 
 
+.. setting:: DOWNLOAD_MAXSIZE
+
+DOWNLOAD_MAXSIZE
+================
```
kmike, line 212: I don't think we need to copy the documentation here, we should just mention that these two standard Scrapy settings are supported.

scrapy_zyte_api/responses.py
```diff
@@ -166,10 +171,67 @@ def from_api_response(cls, api_response: Dict, *, request: Request = None):
...
+    warnsize = request.meta.get("download_warnsize", default_warnsize)
+
+    if "browserHtml" in api_response:
+        expected_size = len(api_response["browserHtml"].encode(_DEFAULT_ENCODING))
+    elif api_response.get("httpResponseHeaders") and api_response.get("httpResponseBody"):
+        expected_size = None
+        for header in api_response.get("httpResponseHeaders"):
+            if header["name"].lower() == "content-length":
+                expected_size = int(header["value"])
+                break
```
kmike, line 211: Why is computing expected size needed? We already got the httpResponseBody, and can check its real size.

scrapy_zyte_api/responses.py
```diff
@@ -166,10 +171,67 @@ def from_api_response(cls, api_response: Dict, *, request: Request = None):
...
+
+        if expected_size is None or (
+            (maxsize and expected_size < maxsize)
+            and (warnsize and expected_size < warnsize)
+        ):
+            expected_size = len(b64decode(api_response.get("httpResponseBody", b"")))
```
kmike, line 217: Is there a way to get size of base64 data without decoding it? Decoding can be costly.

scrapy_zyte_api/responses.py
```diff
@@ -166,10 +171,67 @@ def from_api_response(cls, api_response: Dict, *, request: Request = None):
...
+    api_response: _API_RESPONSE,
+    request: Request,
+    default_maxsize: Optional[int],
+    default_warnsize: Optional[int],
+) -> bool:
+    maxsize = request.meta.get("download_maxsize", default_maxsize)
+    warnsize = request.meta.get("download_warnsize", default_warnsize)
+
+    if "browserHtml" in api_response:
+        expected_size = len(api_response["browserHtml"].encode(_DEFAULT_ENCODING))
```
kmike, line 205: Here while trying to limit the memory using DOWNLOAD_MAXSIZE we might create an additional memory spike, because we create another duplicate of browserHtml in memory, temporarily. 

It seems we need to ensure _response_max_size_exceeded never makes copies of large data from the response.

scrapy_zyte_api/responses.py
wRAR: *.75

wRAR: (assuming no linebreaks or ignoring them)

kmike: It looks fine not to be byte-precise.

scrapy_zyte_api/responses.py
PyExplorer: The idea was to check it first (as this is faster) and if "content-length" is exceed the limit - return without checking the length of the real body.

scrapy_zyte_api/responses.py
kmike: I think you can either use sys.getsizeof (and maybe subtract some fixed overhead Python Unicode objects have), or consider the length of the unicode object instead of the length of the actual data, it could be good enough as well (though worse). Maybe there is some other solution.

PyExplorer: And, we already calculate base64 in two places
https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/responses.py#L197
and 
https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/responses.py#L127
But not sure if this is an issue and we need to fix it the PR.

PyExplorer: Another approach is to calculate the size and decoded/encoded version of the response body here before calling `from_api_response` and send the prepared body to `from_api_response`. In this case we make this expensive calculation only once.

kmike: I was also thinking about moving it down the stack - check the size of the API response received by the client library before the json decoding. But it could make the library less compatible with Scrapy. Let's say you have an existing spider, which uses some download limit. You switch to scrapy-zyte-api for downloads, and maybe also enable data extraction. But the API response is larger than that raw response size. So, the limit becomes more aggressive, and you might drop some pages which were working before.

Because of this, the approach you've taken originally - checking httpResponseBody size and browserHtml size, ignoring everything else (e.g. structured data sizes, or screenshot sizes) makes sense to me.

scrapy_zyte_api/responses.py
PyExplorer: Do you think to remove this check for `content-length` at all?
Actually, I've added this as we consider (and in Scrapy this is also mentioned for DOWNLOAD_MAXSIZE)  to check compressed and decompressed data, and only one way that I found how to check compressed data was to check `content-length`.

scrapy_zyte_api/responses.py
PyExplorer: Ok, let me prepare another version with implementing check here https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/responses.py#L197
and
https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/responses.py#L127

scrapy_zyte_api/responses.py
kmike: Yes, drop it. In Scrapy it's different, because by checking content-length Scrapy can prevent download before it happens. 

For decompression, there is also a special support in Scrapy; it's unrelated to content-length. Scrapy decompresses in chunks, and keeps track of the total size of decompressed data. If the size grows over the limit, an exception is raised, and decompression is stopped. See https://github.com/scrapy/scrapy/blob/6d65708cb7f7b49b72fc17486fecfc1caf62e0af/scrapy/utils/_compression.py#L53. This also looks like something we can't apply here.

PyExplorer: Got it, thanks!

scrapy_zyte_api/responses.py
PyExplorer: @kmike a new version is here.

scrapy_zyte_api/responses.py
```diff
@@ -113,7 +118,9 @@ def _prepare_headers(cls, api_response: Dict[str, Any]):
 
 class ZyteAPITextResponse(ZyteAPIMixin, HtmlResponse):
     @classmethod
-    def from_api_response(cls, api_response: Dict, *, request: Request = None):
+    def from_api_response(
+            cls, api_response: Dict, maxsize: Optional[int], warnsize: Optional[int], *, request: Request = None
```
kmike, line 122: maxsize and warnsize look out of place here. It could make sense to have them here if some work can be prevented, but with the current implementation it seems you can just as easy check len(response.body) after the response is created. 

Because of this, it looks like we can simplify the code by moving the response.body size check to `_download_request` (not even to `_process_response`). This way everything related to max size (reading the settings, showing warnings, checking the body size) would be in the middleware, there is no need to pass parameters around.

kmike (commented):
Looks good overall.

scrapy_zyte_api/handler.py
```diff
@@ -231,7 +234,9 @@ async def _download_request(
         finally:
             self._update_stats(api_params)
 
-        return _process_response(api_response, request, self._cookie_jars)
+        return _process_response(
+            api_response, request, self._cookie_jars, self._default_maxsize, self._default_warnsize
```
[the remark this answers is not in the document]
PyExplorer: Fixed!

PyExplorer:
@kmike please take a look - now the check is in handler_download_request.

scrapy_zyte_api/handler.py
```diff
@@ -23,6 +23,27 @@
...
+) -> bool:
+    if warnsize and body_size > warnsize:
+        logger.warning(
+            f"Actual response size {body_size} larger than "
+            f"download warn size {warnsize} in request {request_url}."
+        )
+
+    if maxsize and body_size > maxsize:
+        logger.warning(
+            f"Cancelling download of {request_url}: actual response size "
```
kmike, line 40: This message seems to be incorrect: we're not cancelling download anymore (as download already happened - we can't cancel it, unlike Scrapy's default download handler), we're dropping the response.

PyExplorer: Ah, of course, that's more correct - fixed!

scrapy_zyte_api/handler.py
```diff
@@ -231,7 +255,18 @@ async def _download_request(
...
+        )
+        if process_response and _body_max_size_exceeded(
+            len(process_response.body),
+            self._default_warnsize,
+            self._default_maxsize,
+            request.url,
+        ):
+            return None
+
+        return process_response
```
kmike, line 269:
```suggestion
        response = _process_response(
            api_response=api_response, request=request, cookie_jars=self._cookie_jars
        )
        if response and _body_max_size_exceeded(
            len(process_response.body),
            self._default_warnsize,
            self._default_maxsize,
            request.url,
        ):
            return None

        return response
```

scrapy_zyte_api/handler.py
```diff
@@ -231,7 +255,18 @@ async def _download_request(
         finally:
             self._update_stats(api_params)
 
-        return _process_response(api_response, request, self._cookie_jars)
+        response = _process_response(
+            api_response=api_response, request=request, cookie_jars=self._cookie_jars
+        )
+        if response and _body_max_size_exceeded(
+            len(process_response.body),
```
kmike, line 262:
```suggestion
            len(response.body),
```

PyExplorer:
@kmike @Gallaecio the PR is ready to be merged. Is it ok to squash+merge it?

PR #228, opened 2024-10-22
Count Zyte API requests from the downloader middleware itself

Gallaecio:


scrapy_zyte_api/_middlewares.py
```diff
@@ -124,29 +125,15 @@ def process_request(self, request, spider):
...
-        download_req_count = sum(
-            [
-                len(slot.transferring)
-                for slot_id, slot in downloader.slots.items()
-                if slot_id.startswith(self._slot_prefix)
-            ]
-        )
-        total_requests = zapi_req_count + download_req_count
-        return total_requests >= self._max_requests
+        self.slot_request(request, spider, force=True)
```
Gallaecio, line 136: Moved after the count to skip this part if the request is going to be ignored anyway.

kmike (approved):

PR #218, opened 2024-09-10
Sync new Zyte API features

Gallaecio:

PR #242, opened 2025-01-16
provider: add serp

Gallaecio:
Subset of https://github.com/scrapy-plugins/scrapy-zyte-api/pull/218 that is a first step towards allowing the `serp` spider from https://github.com/zytedata/zyte-spider-templates to use Serp page objects.


kmike (approved):

wRAR:
We need to add the same thing 5 places, I wonder if it can be made more error-prone :)

wRAR (approved):

PR #248, opened 2025-02-13
Add ZYTE_API_SESSION_MAX_CHECK_FAILURES

Gallaecio:
cc @proway2 @Nykakin

### DIFF
--- a/docs/reference/settings.rst
+++ b/docs/reference/settings.rst
@@ -442,6 +442,20 @@ Default: ``{}``
 overrides of :setting:`ZYTE_API_SESSION_POOL_SIZE` for those pools.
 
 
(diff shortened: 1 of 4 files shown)


proway2:
@Gallaecio sorry for asking this, but am I right assuming that if `ZYTE_API_SESSION_MAX_CHECK_FAILURES = 1` then it will work as before :-)

Gallaecio:
> @Gallaecio sorry for asking this, but am I right assuming that if `ZYTE_API_SESSION_MAX_CHECK_FAILURES = 1` then it will work as before :-)

Exactly, and that is the default value.

kmike (approved):

PR #253, opened 2025-02-28
"zyte-" headers filtered httpResponsebody

GeorgeA92:
"zyte-" headers filtered from httpResponsebody / customHttpRequestHeaders

### DIFF
--- a/scrapy_zyte_api/_params.py
+++ b/scrapy_zyte_api/_params.py
@@ -335,27 +335,37 @@ def _is_safe_header(k, v, /, *, api_params, request):
                     )
                 else:
                     processed_value = processor(decoded_v)
-                    api_params[zapi_request_param] = processed_value
-                    if decoded_v == processed_value:
-                        logger.warning(
-                            f"Request {request} defines header "
-                            f"{decoded_k}. This header has been dropped, "
-                            f"the HTTP API of Zyte API does not support "
-                            f"proxy mode headers, and its value "
-                            f"({decoded_v!r}) has been assigned to the "
-                            f"matching HTTP API request parameter, "
-                            f"{zapi_request_param!r}."
-                        )
+                    if processed_value != _DEFAULT_API_PARAMS[zapi_request_param]:
+                        api_params[zapi_request_param] = processed_value
+                        if decoded_v == processed_value:
+                            logger.warning(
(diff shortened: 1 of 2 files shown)


Gallaecio:
This only drops headers during mapping from `Request.headers`. Should we also detect and drop such headers in `customHttpRequestHeaders`? And if so, should we also automatically map them to HTTP API parameter? Should we have a different behavior when using `zyte_api` instead of `zyte_api_automap`?

GeorgeA92:
@Gallaecio 
> This only drops headers during mapping from `Request.headers`. Should we also detect and drop such headers in `customHttpRequestHeaders`?

As far as I understand `_iter_headers` where check condition updated affects both `Request.headers` and `customHttpRequestHeaders` as `_iter_headers` called on both `_map..` calls here:

https://github.com/scrapy-plugins/scrapy-zyte-api/blob/53071002178817b1a7baa0f48404d1a472c175b4/scrapy_zyte_api/_params.py#L464-L477

and here:
https://github.com/scrapy-plugins/scrapy-zyte-api/blob/53071002178817b1a7baa0f48404d1a472c175b4/scrapy_zyte_api/_params.py#L445-L461

Gallaecio:
Those calls are to map both the `requestHeaders` (browser) and `customHttpRequestHeaders` (HTTP) fields. But if you manually set `customHttpRequestHeaders` on a request, e.g. `"zyte_api_automap": {"customHttpRequestHeaders": […]}`, those will not be parsed, they are taken as is, no automatic header mapping is done, no header dropping, and a warning is logged encouraging the use of `Request.headers` instead.

GeorgeA92:
@Gallaecio 

>  But if you manually set customHttpRequestHeaders on a request, e.g. `"zyte_api_automap": {"customHttpRequestHeaders": […]}`, those will not be parsed, they are taken as is, no automatic header mapping is done, no header dropping, and a warning is logged encouraging the use of Request.headers instead.

On my attempt to reproduce this I got 400 Bad request (while indeed this request passed middleware/handler checks)

Then `_map_custom_http_request_headers` call needs to be moved or called again somewhere on handler side (right before request will be sent) to cover values received  from automap too.

wRAR (approved):

PR #281, opened 2026-04-23
Account for scrapy-poet Zyte API parameters when fingerprinting

AdrianAtZyte:
Before this change, one of the 2 URLs in the example below would be skipped:

```python
from scrapy import Request, Spider
from scrapy_poet import DummyResponse
from zyte_common_items import JobPostingNavigation

class TestSpider(Spider):
    name = "test"
    custom_settings = {
        "ZYTE_API_LOG_REQUESTS": True,
    }

    def start_requests(self):
        for slug in ("foo", "bar"):
            yield Request(
                f"https://toscrape.com/#/{slug}",
                meta={
                    "zyte_api_provider": {
                        "jobPostingNavigationOptions": {"extractFrom": "browserHtml"},
                    },
                },
            )

    async def parse(self, response: DummyResponse, item: JobPostingNavigation):
        yield item
```


scrapy_zyte_api/_request_fingerprinter.py
```diff
@@ -127,29 +129,116 @@ def get_pool(self, request):
...
+                serialized_page_params = (
+                    self._fallback_request_fingerprinter.serialize_page_params(request)
+                )
+                if deps_key is not None:
+                    fingerprint += deps_key
+                if serialized_page_params is not None:
+                    fingerprint += serialized_page_params
+            return hashlib.sha1(fingerprint, usedforsecurity=False).digest()
+
+        def _get_provider_request_fingerprint(self, request: Request) -> bytes | None:
```
wRAR: My only question is whether this makes costly calculations for requests passed to it that are already made for them somewhere else

AdrianAtZyte: I’ve done my best to minimize this now (as of https://github.com/scrapy-plugins/scrapy-zyte-api/pull/281/commits/d5b486321db0287fae6a748ce864ed4c160eadec).

wRAR (approved):

PR #285, opened 2026-06-09
Cover the user of get_retry_request

AdrianAtZyte:


wRAR (approved):

wRAR:
For mypy failures we should pin Twisted==25.5.0 in the tox env and upgrade it separately.

